Issue #2, opened 2016-09-30
[PowerShell] - Make Install-Module automatically install in CurrentUser scope for non admin users


kilasuit:
As per issue https://github.com/PowerShell/PowerShell/issues/2292 - Reported by @ffeldhaus - moving this here as the home for PowerShellGet module for possible further discussion.

This is a feature request. Currently Install-Module gives a warning in Windows PowerShell if it is not run with admin privileges and -Scope CurrentUser is not specified.

As many users just expect Install-Module to work and as the ability to install Modules in the current users scope is one of the key advantages of PowerShell, it would be a great enhancement if the default behaviour of Install-Module would be to automatically install the Module in the current user scope if it is run with non admin privileges.

cc @bmanikm @JKeithB @andschwa

Velocet:
But sometimes it happens that you forget to fire up a shell with admin rights...

So a warning message is a good thing in this case...

Jaykul:
Tt's possible to distinguish between a non-elevated administrator and a non-administrator, right? So perhaps a warning (or the current behavior) for non-elevated administrators, and silently doing the right thing for other users?

bmanikm:
As discussed in https://github.com/PowerShell/PowerShell/issues/2292 I have improved the error message in #4 to detect and throw proper error message in Linux and Mac.

PR #4, opened 2016-10-03
Merging the PowerShellGet changes from PowerShell\PowerShell repo.

bmanikm:
Includes the fix for https://github.com/PowerShell/PowerShell/issues/2286 to provide proper error message when an user is trying to install a module/script as non-root user.
#2

### DIFF
--- a/PowerShellGet/PSModule.psm1
+++ b/PowerShellGet/PSModule.psm1
@@ -88,6 +88,9 @@ $script:PSGetSettings = $null
 $script:MyDocumentsInstalledScriptInfosPath = Microsoft.PowerShell.Management\Join-Path -Path $script:MyDocumentsScriptsPath -ChildPath 'InstalledScriptInfos'
 $script:ProgramFilesInstalledScriptInfosPath = Microsoft.PowerShell.Management\Join-Path -Path $script:ProgramFilesScriptsPath -ChildPath 'InstalledScriptInfos'
 
+$script:IsRunningAsElevated = $true
+$script:IsRunningAsElevatedTested = $false
+
 $script:InstalledScriptInfoFileName = 'InstalledScriptInfo.xml'
 $script:PSGetInstalledScripts = $null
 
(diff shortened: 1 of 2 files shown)


PowerShellGet/PSModule.psm1
```diff
@@ -13721,16 +13761,75 @@ function Copy-Module
...
+                                                    -Force `
+                                                    -ErrorVariable ev `
+                                                    -ErrorAction SilentlyContinue `
+                                                    -WarningAction SilentlyContinue `
+                                                    -Confirm:$false `
+                                                    -WhatIf:$false
+
+        if($ev)
+        {
+            $script:IsRunningAsElevated = $false
```
jianyunt, line 13781: > $script:IsRunningAsElevated = $false [](start = 12, length = 36)

what's the reason we need to set IsRunnignAsElevated to false before throwing? #Closed

PowerShellGet/PSModule.psm1
```diff
@@ -7580,21 +7583,26 @@ function Test-RunningAsElevated
...
-        # This is totally different from Windows where you can know what you can or cannot do with/without admin rights.
-        return $true
+        if($script:IsWindows)
+        {
+            $wid=[System.Security.Principal.WindowsIdentity]::GetCurrent()
+            $prp=new-object System.Security.Principal.WindowsPrincipal($wid)
+            $adm=[System.Security.Principal.WindowsBuiltInRole]::Administrator
+            $script:IsRunningAsElevated = $prp.IsInRole($adm)
+        }
+        elseif($script:IsCoreCLR)
```
jianyunt, line 7595: > elseif($script:IsCoreCLR) [](start = 8, length = 25)

does this mean on CoreCLR we always require admin?  #Closed

PowerShellGet/PSModule.psm1
bmanikm: This will be used for next install/update command to throw proper error/warning prior to downloading. This is used in Test-RunningAsElevated function.

PowerShellGet/PSModule.psm1
bmanikm: Yes, for Linux and Max, we assume that user has the admin rights, $script:IsRunningAsElevated will be updated on first copy error.

jianyunt:
:shipit:

PR #93, opened 2017-02-21
Added credential parameter to subsequent calls of Publish-Module/Script.

dotps1:
This is needed when a module is published that has the RequiredModules attribute in the manifest, and is being published to a Repo that does not have Anonymous access because the required module lookups will fail because there is no read access to repo.

### DIFF
--- a/PowerShellGet/PSModule.psm1
+++ b/PowerShellGet/PSModule.psm1
@@ -792,7 +792,7 @@ function Publish-Module
 
         [Parameter(ValueFromPipelineByPropertyName=$true)]
         [PSCredential]
-        $Credential = $null,
+        $Credential,
 
         [Parameter()] 
         [ValidateSet("1.0")]
@@ -1335,7 +1335,7 @@ function Find-Module
 
         [Parameter(ValueFromPipelineByPropertyName=$true)]
         [PSCredential]
-        $Credential = $null
+        $Credential
     )
 
     Begin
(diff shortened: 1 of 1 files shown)


dotps1:
for the life of me i cannot figure out why the Script tests are failing with:

> Publish-PSArtifactUtility : A parameter cannot be found that matches parameter name 'Credential'.

i will continue to look tonight, but this change needs to happen.

dotps1:
and the issues continue!  https://github.com/OneGet/oneget/issues/270

PowerShellGet/PSModule.psm1
```diff
@@ -792,7 +792,7 @@ function Publish-Module
 
         [Parameter(ValueFromPipelineByPropertyName=$true)]
         [PSCredential]
-        $Credential,
+        $Credential = $null,
```
bmanikm: > = $null [](start = 19, length = 8)

What is the need for specifying the default value? #Closed

PowerShellGet/PSModule.psm1
```diff
@@ -1334,7 +1335,7 @@ function Find-Module
 
         [Parameter(ValueFromPipelineByPropertyName=$true)]
         [PSCredential]
-        $Credential
+        $Credential = $null
```
bmanikm: >$Credential = $null [](start = 8, length = 19)

Please avoid parameters with $null as default values.
 #Closed

dotps1: I have removed the default values.  I swear there used to be a PSScriptAnalyzer rule that was like `PSAvoidUninitalizedParamerterValues`  or something like that where basically if the parameter was not Mandatory, then it needed a default value.  I suppose this would have worked as well: 
`Credential = [PSCredential]::Empty`

Thanks.

bmanikm:
:shipit:

PR #97, opened 2017-03-07
Fixed -f placeholder in ModuleIsNotCatalogSigned

juneb:
In ModuleIsNotCatalogSigned, the placeholder value {0} represents the newest version of the module being installed, which doesn't make sense after the modifier "previously-installed module". 

```
the previously-installed module '{0}' with version '{4}'
```

I changed the placeholder value to {1} which represents the module name.

For example, changed:

```
the previously-installed module '4.0' with version '1.0'
```

To:

```
the previously-installed module 'CoolModule' with version '1.0'
```


bmanikm:
@yutingc Could you please take a look at the two failed test cases related to Update-ModuleManifest cmdlet? 
With the latest changes for Test-ModuleManifest cmdlet in PS Core, we might have to update these two failed test cases.

yutingc:
@bmanikm  Yes, with the new check in Test-ModuleManifest, RootModule and Dependency fields need to point to a valid .psm1, , .cdxml, .xaml or .dll.  We should update the tests accordingly.

bmanikm (approved):

PR #117, opened 2017-04-15
Changed Get-EnvironmentVariable to get the unexpanded version of %path%

hahndorf:
This is in regards to issue #116, updating the machine path environment variable in Install-Script removed any variables in that value.

Running the tests for Install-Script did not cover this (and still don't), all the related tests came out as pending. 

I am not sure this change does not have any side effects. Meaning is there code elsewhere that gets the path variable directly from the registry but needs in expanded rather than un-expanded as it is now.

### DIFF
--- a/PowerShellGet/PSModule.psm1
+++ b/PowerShellGet/PSModule.psm1
@@ -12078,11 +12078,27 @@ function Get-EnvironmentVariable
     }
     elseif ($Target -eq $script:EnvironmentVariableTarget.Machine)
     {
-        $itemPropertyValue = Microsoft.PowerShell.Management\Get-ItemProperty -Path $script:SystemEnvironmentKey -Name $Name -ErrorAction SilentlyContinue
-
-        if($itemPropertyValue)
+        if ($Name -eq "path")
         {
-            return $itemPropertyValue.$Name
+            # if we need the path environment variable, we need it un-expanded, otherwise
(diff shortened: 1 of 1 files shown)


PowerShellGet/PSModule.psm1
```diff
@@ -12078,11 +12078,27 @@ function Get-EnvironmentVariable
...
-        if($itemPropertyValue)
-        {
-            return $itemPropertyValue.$Name
+        if ($Name -eq "path")
+        {
+            # if we need the path environment variable, we need it un-expanded, otherwise
+            # when writing it back, we would loose all the variables like %systemroot% in it.
+            # We use the Win32 API directly using DoNotExpandEnvironmentNames
+            # It is unclear whether any code calling this function for %path% needs the expanded version of %path%
+            # There are currently no tests for this code
```
bmanikm, line 12087: >There are currently no tests for this code [](start = 14, length = 42)

We have few tests for validating PATH updates in PSGetInstallScript.Tests.ps1.

bmanikm:
Thanks @hahndorf for submitting this Pull Request.

---
In reply to: [294278439](https://github.com/PowerShell/PowerShellGet/pull/117#issuecomment-294278439) [](ancestors = 294278439)

bmanikm:
:shipit:

PR #119, opened 2017-04-24
Add "PrivateData" field to ScriptFileInfo

gogbg:
Adding Field "PrivateData" to the ScriptFileInfo. It has the same purpose as the "PrivateData" field in the ModuleManifest -> Storing Custom Script Matadata.

### DIFF
--- a/PowerShellGet/PSModule.psm1
+++ b/PowerShellGet/PSModule.psm1
@@ -8254,12 +8254,19 @@ function Publish-PSArtifactUtility
         }
 
         # Populate the dependencies elements from RequiredModules and RequiredScripts
-        # 
-        $DependentModuleDetails += ValidateAndGet-ScriptDependencies -Repository $Repository `
-                                                                     -DependentScriptInfo $PSScriptInfo `
-                                                                     -CallerPSCmdlet $PSCmdlet `
-                                                                     -Verbose:$VerbosePreference `
-                                                                     -Debug:$DebugPreference
+        #
+        $ValidateAndGetScriptDependencies_Params = @{
+            Repository=$Repository
+            DependentScriptInfo=$PSScriptInfo
+            CallerPSCmdlet=$PSCmdlet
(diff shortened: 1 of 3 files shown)


bmanikm:
@gogbg 
Could you please us know the need for PrivateData in PSScriptInfo? How and where this information will be useful? 
In case of module manifest, PrivateData can be any object ```[-PrivateData <Object>]```  and it has multiple purposes. 

Since, PSScriptInfo is a comment based metadata, PrivateData was not added to PSScriptInfo comment for script files. Please note PSScriptInfo is only used by PowerShellGet. PowerShell runtime doesn't use this info.

gogbg:
In my particular case, I`m developing PS Build module for the purpose similar as msbuild in c#, e.g.
- Module Dependency validation
- Command Reference validation
- Other validations
- Missing module resolution from PSGet Repos

And the specific purpose for which I need to store some metadata for the module/script is:

- Version Control validations, e.g. based on the "FileHash"+"Version for this specific hash" to decide whether the content has been modified or not. So my I`m storing the VersionControl information in the PSPrivateData field.

bmanikm:
Hi @gogbg , 
Could you please elaborate on `"FileHash"+"Version for this specific hash"` for script files? Especially, which file's filehash?

gogbg:
Hi @bmanikm,
In case of script:
Reading from the Script AST, I`m making hash of the param block + dynamic param block + begin block + process block + end block. This way I`m not including the PSScriptInfo, and can change it without affecting the hash

bmanikm:
@gogbg Thanks for the details.
Please add test cases with PrivateData entry for New/Update/Test-ScriptFileInfo cmdlets. 
Ref: https://github.com/PowerShell/PowerShellGet/blob/development/Tests/PSGetPublishScript.Tests.ps1

gogbg:
@bmanikm cool will do it, thanks.

bmanikm:
:shipit:

PR #139, opened 2017-06-12
update readme with the docs edits

TylerLeonhardt:


bmanikm (changes requested):
Consider adding a reference to 
https://msdn.microsoft.com/en-us/powershell/gallery/psget/get_psget_module instead of maintaining the same content at multiple places.

bmanikm (approved):

PR #162, opened 2017-08-24
Added version for REQUIREDSCRIPTS

wooseung-sim:
Added more functionality on 'REQIUREDSCRIPTS' when publishing modules/scripts
It allows user can specify a version of each 'REQIUREDSCRIPTS'. The legacy way that passing only script names are also available.
e.g.> Define 'REQUIREDSCRIPTS' without version (latest version)
.REQUIREDSCRIPTS DummyScript DummyScript2
e.g.>. Define 'REQUIREDSCRIPTS' with specific version
.REQUIREDSCRIPTS DummyScript;2.2.6 DummyScript2;1.0.1

### DIFF
--- a/PowerShellGet/PSModule.psm1
+++ b/PowerShellGet/PSModule.psm1
@@ -7822,20 +7822,24 @@ function ValidateAndGet-ScriptDependencies
                                         Debug = $DebugPreference
                                     }
             $ReqScriptInfo = @{}
-            $scriptName = $requiredScript
-            $scriptVersion = [string]::Empty
 
             if ($PSBoundParameters.ContainsKey('Credential'))
             {
                 $FindScriptArguments.Add('Credential',$Credential)
             }
             
-            if ($scriptName.indexOf(';') -gt 0)
+            if (-not ($scriptName -match '^((?<Name>[^:]+)(:(?<Version>[^:]+))?)$'))
             {
-                $scriptName = $requiredScript.Substring(0, $requiredScript.indexOf(';'))
-                $scriptVersion = $requiredScript.Substring($requiredScript.indexOf(';') + 1, $requiredScript.Length - $requiredScript.indexOf(';') - 1)
+                $message = $LocalizedData.FailedToParseRequiredScripts -f ($scriptName)
+                ThrowError -ExceptionName "System.ArgumentException" `
+                            -ExceptionMessage $message `
+                            -ErrorId "FailedToParseRequiredScripts" `
+                            -CallerPSCmdlet $CallerPSCmdlet `
+                            -ErrorCategory InvalidOperation
             }
 
+            $scriptName = $matches['Name']
+            $scriptVersion = $matches['Version']
             if ($DependentScriptInfo.ExternalScriptDependencies -contains $scriptName)
             {
                 Write-Verbose -Message ($LocalizedData.SkippedScriptDependency -f $scriptName)
--- a/PowerShellGet/PSModule.psm1
+++ b/PowerShellGet/PSModule.psm1
@@ -7828,14 +7828,16 @@ function ValidateAndGet-ScriptDependencies
                 $FindScriptArguments.Add('Credential',$Credential)
             }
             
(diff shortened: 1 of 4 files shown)


bmanikm:
@victor-sim 
Thanks for your contribution. please complete the Contribution License Agreement at https://cla.microsoft.com.
From the current code changes, it looks like versioning requirements are validated only during publish operation. What about during install and update operations?
Please note that side-by-side installation is not supported for scripts. How is the versioning requirements going to work with multiple scripts with different dependency script version requirements?

Please consider creating an issue with the design proposal for supporting version ranges for script dependencies.

wooseung-sim:
@bmanikm 
> From the current code changes, it looks like versioning requirements are validated only during publish operation. What about during install and update operations?

Once it passes requiredscript validation, a script will be packaged with 'RequiredVersion' condition. So nuget should handle this, just like module dependencies.

> Please note that side-by-side installation is not supported for scripts. How is the versioning requirements going to work with multiple scripts with different dependency script version requirements?

 I also considered adopting various options of version, such as min/max/required version support by using NuSpec version formatting or using hashtable format like module dependency.
* '(', ')' - exclusive, '[', ']' - inclusive, '\<MinVersion\>,\<MaxVersion\>'

But I just stopped implementing nuspec version format and make it simple to have only required version, as there's no version conflict resolution for dependencies like you said.
As I didn't feel it worth enough without conflict resolution, I intentionally implemented it as is.

If you feel having nuspec version format is worth enough, I can keep going on this and send another pull request

Thank you.

wooseung-sim:
I've created issue #164

bmanikm:
If you would like to just enable the exact version requirement, please consider `<ScriptName>:[<version>]`. With this, later we can enable other version formats similar to the nuspec's dependency version formatting.

Please add required tests as well.

//Cc @JKeithB @KrishnaV-MSFT

wooseung-sim:
@bmanikm 
Thank you for your comment.
I've pushed another commit which includes, supporting nuspec version string format and unit test.
It doesn't support full variations of nuspec version string types as only 4 types of nuspec version strings are supported on PSGet at packaging time. 
>[1.0]  - RequiredVersion
[1.0,2.0] - Min and Max Version
(,1.0] - Max Version
1.0 - Min Version

PowerShellGet/PSModule.psm1
```diff
@@ -7858,6 +7889,146 @@ function ValidateAndGet-ScriptDependencies
...
 
+function ValidateAndGet-NuspecVersionString
+{
+    param(
+        [Parameter(Mandatory=$true)]
+        [string]
+        $Version
+    )
+
+    $versionPattern = '^((?<MinRule>[\[\(])?((?<MinVersion>[^:\(\[\)\]\,]+))?((?<Comma>[\,])?(?<MaxVersion>[^:\(\[\)\]\,]+)?)?(?<MaxRule>[\]\)])?)$'
```
bmanikm, line 7900: >MinVersion [](start = 48, length = 10)

Should we ensure that extracted version string is a proper version? Users can specify [1112], [ewqew323], [1.1.1.1.1], [1.foo], etc.,?

bmanikm (changes requested):
:clock1:

PowerShellGet/PSModule.psm1
wooseung-sim: ValidateAndGet-NuspecVersionString cmdlet does not validate version string fully. It checks format of the version string.
Such as checking inclusive/exclusive parentheses and/or brackets come together or not, checking existence of version string, detecting invalid or not supported expressions.

And ValidateAndGet-ScriptDependencies cmdlet ensures extracted version string is a proper version by using Find-Script cmdlet, just like it ensure version of dependency modules.

bmanikm: Okay, thanks!

---
In reply to: [135691066](https://github.com/PowerShell/PowerShellGet/pull/162#discussion_r135691066) [](ancestors = 135691066)

bmanikm (approved):
:shipit:

Issue #25, opened 2016-10-17
Enable Publish-Module and Publish-Script cmdlets on PowerShell Core Edition


SteveL-MSFT:
_From @raghushantha on August 8, 2016 18:50_

These cmdlets in PowerShellGet module take dependency on nuget.exe which is windows full sku specific

We need to rework the logic to use Nuget Client APIs so that the publish-\* cmdlets work on CorePowerShell

_Copied from original issue: PowerShell/PowerShell#1678_

pcgeek86:
Any thoughts on prioritization for this? Would love to see the ability to publish modules from PowerShell Core. @bmanikm @SteveL-MSFT

waffleshop:
+1

SteveL-MSFT:
cc @KrishnaV-MSFT

vors:
I'm amazed that it's still an issue.

I want to publish a new version of ZLocation https://ci.appveyor.com/project/vors/zlocation/build/0.2.0.104 and I have a build from a CI windows machine, but I don't have windows to publish it.
Current behavior is not only nuget, but essentially a 4+ Gb image of the windows OS.

SteveL-MSFT:
I just did a simple test removing the check for $IsCoreCLR and also defining $IsNanoServer = $false (the current check is incorrect and should be changed to [this](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/CoreCLR/CorePsPlatform.cs#L78)).  Some problem with `Get-DSCResource` so I made `Get-ExportedDscResources` just return an empty array.  After this, `publish-module` worked just fine on PSCore6 Beta.5.

If I find some time from PSCore6 issues, maybe I'll submit a PR for this.

SteveL-MSFT:
@vors unfortunately my fix only makes it work on PSCore6 on Windows, still not supported on non-Windows due to dependency on nuget.exe

SteveL-MSFT:
Looks like my change only enables it for PSCore on Windows, so not a complete fix for non-Windows platforms.  No longer working on this.

pcgeek86:
Is someone else taking ownership of this @SteveL-MSFT? What does this mean for the first launch version of PowerShell Core Edition?

SteveL-MSFT:
@KrishnaV-MSFT can you comment on whether we will get support by the time 6.0.0 is done?

bmanikm:
[...]

PR #163, opened 2017-08-24
enable publish-module to work on PSCore6

SteveL-MSFT:
This is only a partial fix as it works on Windows PSCore6, but due to dependency on nuget.exe it doesn't work on non-Windows

Related https://github.com/PowerShell/PowerShellGet/issues/25


bmanikm:
We need to use [dotnet commands](https://docs.microsoft.com/en-us/nuget/tools/dotnet-commands) to support Publish-Module cmdlet on non-Windows platforms.

SteveL-MSFT:
@bmanikm yeah, I noticed that after I tried testing my changes on Linux so I'm only allowing it to work on PSCore6 on Windows.  Slightly better, but some customers are asking for Linux/Mac support.

SteveL-MSFT:
Unfortunately the failing tests don't fail on my desktop so I'm not sure how much time I'm going to spend on this particularly since it doesn't enable it on Linux/Mac.  In any case, I would suggest taking the NanoServer detection code as the current code is just wrong and incorrectly identifies PSCore6 on Windows as Nano.

PR #179, opened 2017-10-17
Ensure MyDocumentsPSPath path is correct

lwsrbrts:
Ensure that the user's My Documents PowerShell path for modules is
correct and not assumed to be under
C:\Users\[user]\Documents\PowerShell\


lwsrbrts:
First actual PR so please let me know if I've dropped a ball somewhere. This PR is in relation to the Issue I filed a few days ago.

#174

bmanikm (approved):

Issue #208, opened 2017-12-14
Find-Module fails when module has a prerelease version


splatteredbits:
When a module (or any module) in the PowerShell Gallery has a prerelease version (like Pester currently does), `Find-Module` fails with a terminating exception and returns nothing.

## Expected Behavior
I expect `Find-Module` to work when searching for a module that has a prerelease version. 

## Current Behavior
(Note: you get the same error with no parameters.)
```
> Find-Package -Name 'Pester'
Find-Package : A parameter cannot be found that matches parameter name 'AllowPrereleaseVersions'.
At C:\projects\whiskey\Whiskey\PowerShellGet\PSModule.psm1:1609 char:34
+         PackageManagement\Find-Package @PSBoundParameters | Microsoft ...
+                                        ~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Find-Package], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Microsoft.PowerShell.PackageManagement.Cmdlets.FindPackage

## Possible Solution
Stop passing the `AllowPrereleaseVersions` flag to the `Find-Package` cmdlet. They get returned by default and `Find-Module` is already filtering them out:

```
                if ($psgetItemInfo.AdditionalMetadata -and $psgetItemInfo.AdditionalMetadata.IsPrerelease -eq 'false')
                {
                    $psgetItemInfo
                }
```

## Steps to Reproduce (for bugs)
1. Import PackageManagement 1.1.7
2. Import PowerShellGet 1.6.0
3. Run `Find-Module`.
4. Note the error.

## Context
We have a build tool that auto-installs local copies of modules. We can no longer install Pester because it just released a [prerelease version](https://www.powershellgallery.com/packages/Pester/4.2.0-alpha2).

## Your Environment

```PowerShell

> $PSVersionTable
Name                           Value                                                                                 
----                           -----                                                                                 
PSVersion                      5.0.10586.117                                                                         
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                               
BuildVersion                   10.0.10586.117                                                                        
CLRVersion                     4.0.30319.42000                                                                       
WSManStackVersion              3.0                                                                                   
PSRemotingProtocolVersion      2.3                                                                                   
SerializationVersion           1.1.0.1   

> Get-Module
ModuleType Version    Name                                ExportedCommands                                           
---------- -------    ----                                ----------------                                           
[...]

PR #209, opened 2017-12-14
Fix Issue 208

splatteredbits:
I would have created a test fixture, but didn't know how to insert prerelease versions in the test instance of the gallery you're using. If someone wants to show me how to do that, I'll do it.

Also, should this go to development or master? I'd like to see a 1.6.1 version get released with this fix.


bmanikm:
I have provided the required steps to resolve https://github.com/PowerShell/PowerShellGet/issues/208.

PR #255, opened 2018-04-05
Fix#225/add proxy to publish functions

Benny1007:


bmanikm:
Thanks @Benny1007! Please let us know whether the Publish-Module cmdlet is working fine with the proxy settings.
I think, we need to execute the below commands to ensure that publish operation works fine with the proxy settings, right?
```
nuget.exe config -set http_proxy=http://my.proxy.address:port
nuget.exe config -set http_proxy.user=mydomain\myUserName
nuget.exe config -set http_proxy.password=mySuperSecretPassword
```

bmanikm:
@Benny1007 Closing this PR for now. Please feel free to reactivate this PR when above https://github.com/PowerShell/PowerShellGet/pull/255#issuecomment-379033102 gets resolved.

PR #258, opened 2018-04-06
Allow Pester / PSReadline installation when signed by non-Microsoft certificate

brywang-msft:
First assumption is that Pester and PSReadline will be signed by a non-Microsoft certificate soon. So the scenario becomes whitelisting installation of non-Microsoft signed Pester / PSReadline over Microsoft signed Pester / PSReadline.

A separate Whitelist.json is a possibility but currently veto'd, so a hardcoded list in PSGet code is chosen instead.

The publisher check test only works on full CLR + current PS versions due to reliance on signing during runtime.


Tests/PSGetPublishModule.Tests.ps1
```diff
@@ -55,11 +55,61 @@ function SuiteSetup {
...
+            if (-not (Test-Path -Path $signedPesterDestination)) {
+                $null = New-Item -Path $signedPesterDestination -ItemType Directory
+            }
+            Copy-Item -Path $signedPester -Destination $signedPesterDestination -Recurse -Force
+        }
+        
+        Copy-Item -Path $pesterv1 -Destination $pesterDestination -Recurse -Force
+        Copy-Item -Path $pesterv2 -Destination $pesterDestination -Recurse -Force
+
+        $csCert = Get-CodeSigningCert
```
brywang-msft: Hm, this block doesn't seem to be working... thoughts? #Closed

PowerShellGet/PSGet.Resource.psd1
```diff
@@ -254,6 +254,7 @@ ConvertFrom-StringData @'
         RequiredScriptVersoinFormat=<ScriptName>, <ScriptName>:<MinimumVersion>, <ScriptName>:[<RequiredVersion>], <ScriptName>:[<MinimumVersion>,<MaximumVersion>], <ScriptName>:[,<MaximumVersion>]
         FailedToParseRequiredScripts=Cannot parse REQUIREDSCRIPTS '{0}'. Acceptable formats are: '{1}'.
         FailedToParseRequiredScriptsVersion=Version format error: {0}, '{1}'. Acceptable formats are: '{2}'. 
+        PublishersMismatchAsWarning=A Microsoft-signed module named '{0}' with version '{1}' that was previously installed conflicts with the new module '{2}' from publisher '{3}' with version '{4}'. Installing the new module may result in system instability.
```
edyoung: "May cause instability" sounds a bit dramatic. How about:
Module '{0}' version '{1}' published by '{2}' will be superceded by version '{3}' published by '{4}'. If you do not trust the new publisher, uninstall the module.

Tests/PSGetPublishModule.Tests.ps1
edyoung: It looks like Get-CodeSigningCert will return a suitable cert if that is already available in your cert: store but will just return $null if there isn't one present.

PowerShellGet/PSGet.Resource.psd1
brywang-msft: Same text as ```PublishersMismatch``` but without the ```Add -SkipPublisherCheck``` part. Should that text be changed as well?

Tests/PSGetPublishModule.Tests.ps1
brywang-msft: I think one should exist from a previous test, but if not, ```Create-CodeSigningCert``` should be creating one. I also didn't expect WMF 4/5 to have the PKI module so I'll have to spin up some VMs to look at this, partly so I don't have to mess with my own cert store.

PowerShellGet/private/functions/Validate-ModuleAuthenticodeSignature.ps1
```diff
@@ -1,3 +1,9 @@
+# Modules allowed to install non-Microsoft signed modules over Microsoft signed modules
+$script:WhitelistedModules = @{
+    "Pester"     = $true
+    "PSReadline" = $true
+}
+
```
bmanikm: Please consider moving this var to ...\PowerShellGet\private\modulefile\PartOne.ps1 #Closed

Tests/PSGetPublishModule.Tests.ps1
bmanikm: Please take a look at the catalog singing verification tests in ...\Tests\Pester.CatalogSignedModules.Tests.ps1.

---
In reply to: [179861148](https://github.com/PowerShell/PowerShellGet/pull/258#discussion_r179861148) [](ancestors = 179861148)

Tests/PSGetPublishModule.Tests.ps1
```diff
@@ -1027,6 +1078,56 @@ Describe PowerShell.PSGet.PublishModuleTests -Tags 'BVT','InnerLoop' {
         Assert ($itemInfo.Includes.RoleCapability -contains 'Lev2Maintenance') "Publish-Module was not able to populate the RoleCapability Names: $($itemInfo.Includes.RoleCapability)"
     } `
     -Skip:$($PSCulture -ne 'en-US')
+
+    It 'InstallNonMsSignedModuleOverMsSignedModule' {
```
bmanikm: >InstallNonMsSignedModuleOverMsSignedModule [](start = 8, length = 42)

Please consider moving this test to Pester.CatalogSignedModules.Tests.ps1 file. #Closed

bmanikm:
> 

Please consider creating these two psd1 files in the test setup using New-ModuleManifest cmdlet. #Closed

---
Refers to: Tests/TestModules/PesterTemp/99.99.99.98/Pester.psd1:247 in 8424265. [](commit_id = 8424265ebf26a08470e933d625e6820b497447ae, deletion_comment = False)

PowerShellGet/private/modulefile/PartOne.ps1
```diff
@@ -329,6 +329,12 @@ $script:PackageManagementUpdateModuleMessageResolverScriptBlock =  {
                                                 }
                                             }
 
+# Modules allowed to install non-Microsoft signed modules over Microsoft signed modules
+$script:WhitelistedModules = @{
+    "Pester"     = $true
+    "PSReadline" = $true
+}
```
bmanikm, line 336: What is the purpose of dictionary here? Why array of whitelisted module names is not enough?

Tests/PSGetPublishModule.Tests.ps1
```diff
@@ -77,6 +79,7 @@ function SuiteCleanup {
 
     RemoveItem $script:PSGalleryRepoPath
     RemoveItem $script:TempModulesPath
+    RemoveItem $script:TestPSModulePath
```
bmanikm: >RemoveItem $script:TestPSModulePath [](start = 4, length = 35)

Please revert the changes in this file as you have moved the new test to a different file. #Closed

Tests/PSGetInstallModule.Tests.ps1
```diff
@@ -823,6 +888,61 @@ Describe PowerShell.PSGet.InstallModuleTests -Tags 'BVT','InnerLoop' {
...
+    It 'InstallNonMsSignedModuleOverMsSignedModule' {
+        $pesterRoot = Join-Path -Path $script:TempModulesPath -ChildPath "Pester"
+        $v1Path = Join-Path -Path $pesterRoot -ChildPath "99.99.99.98"
+        $v2Path = Join-Path -Path $pesterRoot -ChildPath "99.99.99.99"
+        # Publish signed modules
+        Publish-Module -Path $v1Path -Repository $script:localGalleryName
+        Publish-Module -Path $v2Path -Repository $script:localGalleryName
+        $oldPSModulePath = $env:PSModulePath
+        $env:PSModulePath = $script:TestPSModulePath
+        try {
```
bmanikm, line 907: Why do we need to change the value of $env:PSModulePath?

It is required to reset the value of $env:PSModulePath when there is any assertion/should failure after v1 installation.

This is the root cause for the CI test failures. #Closed

brywang-msft: $env:PSModulePath needs to be changed so that the highest version is the Microsoft-signed version of Pester, otherwise the test scenario is broken.

PowerShellGet/PSGet.Resource.psd1
edyoung: I think we should change this one and leave the other. I'm not wild about the wording there either, but there it's an unexpected condition - here we're on an expected path and we don't want to freak people out

PowerShellGet/private/modulefile/PartOne.ps1
edyoung: Either would work IMO

edyoung (approved):
Thanks!

bmanikm (approved):
:shipit:

PR #260, opened 2018-04-11
Fix for Get-InstalledScript -RequiredVersion doesn't work when versions have a leading 0

bmanikm:
Additional changes:
- Updated Readme.md to include additional steps for running tests with the merged psmodule.psm1 file.
- Fixed a typo in the verbose message.


edyoung:
Thanks!

PR #263, opened 2018-04-18
Alerickson positional path

alerickson:
Added positional parameter to path.


edyoung:
This contains more changes than expected. I think what's happened is that your branch is a child of the development branch, and this is a merge to master - so it is pulling along all the other changes in the development branch too. I think it's best to abandon this pull request, then make a new one to merge to development instead- that should just show the small change you made.

alerickson:
Oh sorry about that!  The default was to merge to development but I thought
I should change that to master.  I'll delete and create a new pull request.

PR #266, opened 2018-04-25
Alerickson positional path

alerickson:
Added a positional parameter to path in Save-Script.
Added corresponding tests for Save-Script changes and Save-Module changes.


Tests/PSGetInstallModule.Tests.ps1
```diff
@@ -1348,6 +1348,32 @@ Describe PowerShell.PSGet.InstallModuleTests.P1 -Tags 'P1','OuterLoop' {
...
+    #
+    # Action: Find-Module -Name ContosoClient | Save-Module 
+    #
+    # Expected Result: Should save the module
+    #
+    It SaveModuleWithPathPositionalParameter {
+        Install-Module ContosoClient
+        $ModuleName = "ContosoClient"
+
+        $res1 = Find-Module -Name $ModuleName -RequiredVersion "2.5"
```
edyoung, line 1362: We can simplify these tests to make it a bit clearer what we are actually testing. 

We don't need to call Find-Module at all, and I don't really see why we need to do Install-Module either. 
Just

Save-Module $ModuleName $script:TempPath
AssertTrue Test-Path $script:TempPath\$ModuleName

Then we are validating that Save-Module put the files in the path we gave it, which is all we really wanted.

Tests/PSGetInstallScript.Tests.ps1
```diff
@@ -1766,4 +1766,30 @@ Describe PowerShell.PSGet.InstallScriptTests.P1 -Tags 'P1','OuterLoop' {
...
+    #
+    # Action: Find-Script -Name ModuleWithDependencies1 | Save-Script
+    #
+    # Expected Result: Should save the script
+    #
+    It SaveScriptWithPathPositionalParameter {
+        $ScriptName = 'Fabrikam-ClientScript'
+        Install-Script -Name $ScriptName
+
+        $res1 = Get-InstalledScript -Name $ScriptName -RequiredVersion "2.5"
```
edyoung, line 1780: same comment for this script as above.

edyoung:
Thanks Amber! I have a couple minor comments, but this is 99% there.

Also We shouldn't be checking in the testresultscurrent.xml file, please remove that.

edyoung:
oops, clicked close by mistake

edyoung (commented):
Sorry, I forgot to hit 'submit review' for these comments so you didn't see them. Since the change is merged now don't worry too much about these minor comments, but you could opportunistically fix them in a future update

Tests/PSGetInstallModule.Tests.ps1
alerickson: Yup, I can definitely clean up the code in the near future.  There is a lot of extraneous information/redundant code.  I was trying to keep the style similar to the other tests, but I think for the sake of clarity it's best to just clean up and simplify.

PR #272, opened 2018-05-09
Publisher verification

### LINKED ISSUE #160: Two problems with module publisher verification
Function in question:  Get-AuthenticodePublisher (currently on line 14661 in the dev branch: https://github.com/PowerShell/PowerShellGet/blob/development/PowerShellGet/PSModule.psm1#L14661)

First problem is simple to fix:  It is apparently possible to have two identical certificates installed in the same store.  I don't know how, but it causes trouble with PowerShellGet, which assumes that Publisher is going to be a single string rather than an array.  This can be solved by using `Select-Object -First 1` in the assignment to `$rootCertificationAuthority`.

Example of current output:

```
PS C:\Users\dwyatt\Documents\WindowsPowerShell\Modules\ProtectedData\4.1.3>> dir Cert:\LocalMachine\Root\056*


   PSParentPath: Microsoft.PowerShell.Security\Certificate::LocalMachine\Root

Thumbprint                                Subject                                                                                                                                                                                                                                             
----------                                -------                                                                                                                                                                                                                                             
0563B8630D62D75ABBC8AB1E4BDFB5A899B24D43  CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US                                                                                                                                                                           
0563B8630D62D75ABBC8AB1E4BDFB5A899B24D43  CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US 

update-module protecteddata -Force
PackageManagement\Install-Package : Authenticode issuer 'System.Object[]' of the new module 'ProtectedData' with version '4.1.3' is not matching with the authenticode issuer 'System.Object[]' of the previously-installed module 'ProtectedData' with version '4.1.3'. If you still want to install or update, use
-SkipPublisherCheck parameter.
At C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\1.1.3.2\PSModule.psm1:2147 char:20
+ ...           $sid = PackageManagement\Install-Package @PSBoundParameters
+                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (Microsoft.Power....InstallPackage:InstallPackage) [Install-Package], Exception
    + FullyQualifiedErrorId : AuthenticodeIssuerMismatch,Validate-ModuleAuthenticodeSignature,Microsoft.PowerShell.PackageManagement.Cmdlets.InstallPackage
```

The second problem is about design.  The way Get-AuthenticodePublisher is currently written, it walks up the certificate chain, and winds up returning the DigiCert root CA as the publisher of my module (because my code-signing cert comes from them.)  Doesn't that mean that anyone else who happens to have a DigiCert code-signing CA will also be treated as the "same" publisher by PowerShellGet, if they happened to somehow upload a version of my module?  Doesn't that defeat the purpose of having a publisher check in the first place?  Shouldn't it really be treating the subject of my code-signing cert as the publisher, while also verifying that the certificate chain is trusted (not a self-signed cert / etc.)

alerickson:
Resolves #160


Tests/Pester.CatalogSignedModules.Tests.ps1
```diff
@@ -348,6 +431,74 @@ Describe 'Test PowerShellGet\Update-Module cmdlet with catalog signed modules' -
     }
 }
 
+Describe 'Test Install-Module and Update-Module for catalog signed test modules' -tags 'P2','InnerLoop' {
```
bmanikm: >P2 [](start = 88, length = 2)

Please change the priority tag to BVT #Closed

Tests/Pester.CatalogSignedModules.Tests.ps1
```diff
@@ -156,6 +157,88 @@ function SuiteSetup {
...
+    # If TestModule v4 exists, uninstall
+    if (Get-Module TestModule -ListAvailable | Where-Object { $_.Version -eq '99.99.99.99' }) {
+        $moduleBase = (Get-Module TestModule -ListAvailable | Where-Object { $_.Version -eq '99.99.99.99' }).ModuleBase
+        $null = Remove-Item -Path $moduleBase -Force -Recurse
+        if (Get-Module TestModule -ListAvailable | Where-Object { $_.Version -eq '99.99.99.99' }) {
+            Write-Error "Failed to uninstall v4"
+        }
+    }       
+
+    RemoveItem (Join-Path -Path $ProgramFilesModulesPath -ChildPath "TestModule")
```
bmanikm: >RemoveItem (Join-Path -Path $ProgramFilesModulesPath -ChildPath "TestModule") [](start = 4, length = 77)

This handles cleanup of removing all versions of TestModule, so above logic for removing the v1 and v4 versions of TestModule can be removed. #Closed

Tests/Pester.CatalogSignedModules.Tests.ps1
```diff
@@ -348,6 +431,74 @@ Describe 'Test PowerShellGet\Update-Module cmdlet with catalog signed modules' -
...
+      
+    It 'Install new non-MS signed module over current non-MS signed module' {
+        $TestModuleRoot = Join-Path -Path $script:TempModulesPath -ChildPath "TestModule"
+        $TestModulev1Path = Join-Path -Path $TestModuleRoot -ChildPath "99.99.99.96"  
+        $TestModulev2Path = Join-Path -Path $TestModuleRoot -ChildPath "99.99.99.97"  
+        $TestModulev3Path = Join-Path -Path $TestModuleRoot -ChildPath "99.99.99.98"  
+        $TestModulev4Path = Join-Path -Path $TestModuleRoot -ChildPath "99.99.99.99"
+      
+        # Publish and install v1 signed modules
+        Publish-Module -Path $TestModulev1Path -Repository $script:localGalleryName -Force
```
bmanikm: >Publish-Module -Path $TestModulev1Path -Repository $script:localGalleryName -Force [](start = 8, length = 82)

Consider moving Publish-Module operations in this test to SuiteSetup2 #Closed

Tests/Pester.CatalogSignedModules.Tests.ps1
```diff
@@ -156,6 +157,88 @@ function SuiteSetup {
...
+        $null = Set-AuthenticodeSignature -FilePath (Join-Path -Path $TestModulev3Destination -ChildPath "TestModule.psd1") -Certificate $bobCert
+        $null = Set-AuthenticodeSignature -FilePath (Join-Path -Path $TestModulev4Destination -ChildPath "TestModule.psd1") -Certificate $alice1Cert
+    }
+}
+
+
+function SuiteCleanup {
+
+    # Import the PowerShellGet provider to reload the repositories.
+    $null = Import-PackageProvider -Name PowerShellGet -Force
```
bmanikm: We also need to restore the previous snapshot of registered repositories prior to this line. #Closed

bmanikm (approved):
:shipit:

Issue #182, opened 2017-10-26
Update-ModuleManifest reverts back CompanyName and FunctionsToExport to default


mmattes:
When using the following command `Update-ModuleManifest -Path $ManifestPath -ModuleVersion $NewVersion` the CompanyName and FunctionsToExport parameter gets set back to default

## Expected Behavior
`Update-ModuleManifest -Path $ManifestPath -ModuleVersion $NewVersion` should only update the ModuleVersion and should not touch any other value

## Current Behavior
`Update-ModuleManifest -Path $ManifestPath -ModuleVersion $NewVersion` also sets CompanyName from `# CompanyName = ''` to `CompanyName = 'Unkown'` and `FunctionsToExport = '*'` to `FunctionsToExport = @()`

## Steps to Reproduce (for bugs)

1. Create a Module.psd1 and set the two parameters as shown below
```
# CompanyName = ''
FunctionsToExport = '*'
```
2. Run `Update-ModuleManifest -Path Module.psd1 -ModuleVersion "1.1.1.0"`
3. The Parameters in the Module.psd1 are now 
```
CompanyName = 'Unkown'
FunctionsToExport = @()
```

## Context
I am trying to setup a deploy pipepline for PsGallery which automatically updates the version number of my module 

## Your Environment

```PowerShell

> $PSVersionTable
Name                           Value                                                                                                                                             
----                           -----                                                                                                                                             
PSVersion                      6.0.0-beta.9                                                                                                                                      
PSEdition                      Core                                                                                                                                              
GitCommitId                    v6.0.0-beta.9                                                                                                                                     
OS                             Linux 4.4.0-53-generic #74-Ubuntu SMP Fri Dec 2 15:59:10 UTC 2016                                                                                 
Platform                       Unix                                                                                                                                              
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                           
PSRemotingProtocolVersion      2.3                                                                                                                                               
SerializationVersion           1.1.0.1                                                                                                                                           
WSManStackVersion              3.0     

> Get-Module
ModuleType Version    Name                                ExportedCommands                                                                                                       
---------- -------    ----                                ----------------                                                                                                       
Script     0.1.11.0   Invoke-Install                                                                                                                                             
Manifest   3.1.0.0    Microsoft.PowerShell.Management     {Add-Content, Clear-Content, Clear-Item, Clear-ItemProperty...}                                                        
[...]

PR #284, opened 2018-05-21
Modified Update-ModuleManifest so that params will not reset to default values

alerickson:
Parameters modified:  
- CompanyName
- FunctionsToExport
- CmdletsToExport 
- AliasesToExport  
- VariablesToExport 
- DscResourcesToExport

### DIFF
--- a/Tests/PSGetUpdateModuleManifest.Tests.ps1
+++ b/Tests/PSGetUpdateModuleManifest.Tests.ps1
@@ -35,8 +35,8 @@ Describe PowerShell.PSGet.UpdateModuleManifest -Tags 'BVT','InnerLoop' {
     AfterEach {
         RemoveItem "$script:TempModulesPath\*"
     }
-
-
+    
+    
     # Purpose: Validate Update-ModuleManifest will keep the properties the same as with the original manifest using PowerShellGet module as test.
     #
     # Action:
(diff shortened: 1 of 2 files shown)


Tests/PSGetUpdateModuleManifest.Tests.ps1
```diff
@@ -113,6 +113,31 @@ Describe PowerShell.PSGet.UpdateModuleManifest -Tags 'BVT','InnerLoop' {
...
 
+    # Purpose: Validate Update-ModuleManifest will not reset original parameter values to default values
+    #
+    # Action:
+    #      Update-ModuleManifest -Path [Path] 
+    #
+    # Expected Result: The updated manifest should have the same proerty values.
+    #
+    It UpdateModuleManifestWithNoAdditionalParameters2 {    
+        New-ModuleManifest -Path $script:testManifestPath -ModuleVersion '1.0' -FunctionsToExport '*' -CmdletsToExport '*' -AliasesToExport '*' -VariablesToExport '*' -DscResourcesToExport '*'
```
bmanikm: >-DscResourcesToExport '*' [](start = 167, length = 25)

This test failed on PS 4 as -DscResourcesToExport is not supported on PS 4.0 and 3.0, so conditionally add this param here.

Tests/PSGetUpdateModuleManifest.Tests.ps1
```diff
@@ -113,6 +113,31 @@ Describe PowerShell.PSGet.UpdateModuleManifest -Tags 'BVT','InnerLoop' {
...
+
+        #Edit company name from 'Unknown' to ''
+        (get-content $script:testManifestPath) | foreach-object {$_ -replace 'Unknown', ''} | set-content $script:testManifestPath
+        $editedModuleInfo = Test-ModuleManifest -Path $script:testManifestPath
+
+        Update-ModuleManifest -path $script:testManifestPath -ModuleVersion '2.0'
+        #$updatedModuleInfo = Test-ModuleManifest -Path $script:testManifestPath
```
bmanikm: >#$updatedModuleInfo = Test-ModuleManifest -Path $script:testManifestPath [](start = 8, length = 72)

It looks like this line needs to be uncommented, right?

Tests/PSGetUpdateModuleManifest.Tests.ps1
```diff
@@ -113,6 +113,31 @@ Describe PowerShell.PSGet.UpdateModuleManifest -Tags 'BVT','InnerLoop' {
...
+
+        $expectedLength = 5
+        $text = @(Get-Content -Path $script:testManifestPath | Select-String "\*")
+
+        AssertEquals $updatedModuleInfo.CompanyName $editedModuleInfo.CompanyName "Company name should be $expectedCompanyName"
+        AssertEquals $($text.length) $expectedLength "Number of wildcards should be $expectedLength"
+    } `
+    -Skip:$($IsWindows -eq $False)
```
bmanikm: Anything blocking us from enabling this test on non-Windows platforms?

alerickson: There is nothing I know of that is, but other similar tests in this file had that condition, so I thought I would keep in line with those tests.  I can certainly take the condition out.

Tests/PSGetUpdateModuleManifest.Tests.ps1
alerickson: Yes, thank you!

Tests/PSGetUpdateModuleManifest.Tests.ps1
alerickson: Okay, thanks.  I'll add that check.

bmanikm (approved):

alerickson:
Closes #182

edyoung:
thanks!

PR #305, opened 2018-07-25
Change version variables to add build number in PublishModuleWithExternalDependencies test

alerickson:


Tests/PSGetPublishModule.Tests.ps1
```diff
@@ -2033,7 +2033,7 @@ Describe PowerShell.PSGet.PublishModuleTests.P2 -Tags 'P2','OuterLoop' {
         try
         {
             # No warning during Install and Update as externally managed modules are already installed.
-            $version = '1.0'
+            $version = '1.0.0'
```
bmanikm, line 2036: >$version = '1.0.0' [](start = 12, length = 18)

As we discussed offline, please ensure that the module version is 3-part instead of 2-part so that the version validation is correct and works fine on all the supported platforms. #Closed

bmanikm (changes requested):
:clock1:

bmanikm (approved):
:shipit:

PR #315, opened 2018-08-22
Change default scope in Install-Module and Install-Script

alerickson:


PowerShellGet/public/psgetfunctions/Install-Module.ps1
```diff
@@ -91,16 +91,15 @@ function Install-Module
     {
         Get-PSGalleryApiAvailability -Repository $Repository
 
-        if(-not (Test-RunningAsElevated) -and ($Scope -ne "CurrentUser"))
+        # If user has elevated privileges and no scope is specified, default installation will be to all users,
+        # Otherwise default installation will be to current user.
+        if (-not ($Scope)) 
         {
-            # Throw an error when Install-Module is used as a non-admin user and '-Scope CurrentUser' is not specified
-            $message = $LocalizedData.InstallModuleNeedsCurrentUserScopeParameterForNonAdminUser -f @($script:programFilesModulesPath, $script:MyDocumentsModulesPath)
```
edyoung: Consider: when the user is not an admin and they explicitly specify  -Scope AllUsers shouldn't they see this message?

Tests/PowerShellGet.Tests.ps1
```diff
@@ -125,14 +125,37 @@ Describe "PowerShellGet - Module tests" -tags "Feature" {
         $psgetModuleInfo.Repository | Should Be $RepositoryName
     }
 
-    It "Should install a module correctly to the required location with CurrentUser scope" {
+    It "Should install a module correctly to the required location with CurrentUser scope and no elevated privileges" {
```
edyoung: do these tests always run without elevated privileges, or just with the default privileges of how you start the shell? ie if I run these tests in an Administrator window won't they fail?

edyoung (commented):
Thanks!

PowerShellGet/public/psgetfunctions/Install-Module.ps1
alerickson: They still see the error message because this function calls PackageManagement\Install-Module, which has the same error message.  I don't think it's necessary to keep it in there unless there's a specific situation where it's advantageous to throw the error early on.

Tests/PowerShellGet.Tests.ps1
alerickson: The tests will run based off of the default privileges of the shell.  The previous install-module/install-script tests were already split up into "Module tests"  and "Module tests (Admin)" (the latter is tagged with "RequireAdminOnWindows").  Running without privileges causes all the non-elevated privilege tests to pass and vice versa.

edyoung:
Lgtm

PowerShellGet/public/psgetfunctions/Install-Module.ps1
```diff
@@ -91,16 +91,15 @@ function Install-Module
...
-
-            ThrowError -ExceptionName "System.ArgumentException" `
-                        -ExceptionMessage $message `
-                        -ErrorId "InstallModuleNeedsCurrentUserScopeParameterForNonAdminUser" `
-                        -CallerPSCmdlet $PSCmdlet `
-                        -ErrorCategory InvalidArgument
+            $Scope = "CurrentUser"
+            if(Test-RunningAsElevated)
```
bmanikm: >if(Test-RunningAsElevated) [](start = 12, length = 26)

On non-Windows OS platforms, this function returns true as there is no straight forward way to identify whether PS is running as admin. I think, the default scope the non-Windows platforms should be CurrentUser.

Tests/PowerShellGet.Tests.ps1
```diff
@@ -125,14 +125,37 @@ Describe "PowerShellGet - Module tests" -tags "Feature" {
...
+        $installedModuleInfo | Should Be $null
+    }
+
+    It "Should install a module correctly to the required location with default CurrentUser scope and no elevated privileges" {
+        Install-Module -Name $ContosoServer -Repository $RepositoryName
+        $installedModuleInfo = Get-InstalledModule -Name $ContosoServer
+
+        $installedModuleInfo | Should Not Be $null
+        $installedModuleInfo.Name | Should Be $ContosoServer
+        $installedModuleInfo.InstalledLocation.StartsWith($script:MyDocumentsModulesPath, [System.StringComparison]::OrdinalIgnoreCase) | Should Be $true
```
bmanikm: This tests in this file gets executed only in the daily test pass and it is sanity tests file for PS core builds.
In PowerShellGet daily test runs, all the tests in this file are executed with admin privileges.

PowerShellGet/public/psgetfunctions/Install-Module.ps1
```diff
@@ -91,16 +91,15 @@ function Install-Module
...
         {
-            # Throw an error when Install-Module is used as a non-admin user and '-Scope CurrentUser' is not specified
-            $message = $LocalizedData.InstallModuleNeedsCurrentUserScopeParameterForNonAdminUser -f @($script:programFilesModulesPath, $script:MyDocumentsModulesPath)
-
-            ThrowError -ExceptionName "System.ArgumentException" `
-                        -ExceptionMessage $message `
-                        -ErrorId "InstallModuleNeedsCurrentUserScopeParameterForNonAdminUser" `
-                        -CallerPSCmdlet $PSCmdlet `
-                        -ErrorCategory InvalidArgument
+            $Scope = "CurrentUser"
```
bmanikm: >$Scope = "CurrentUser" [](start = 12, length = 22)

It is also required to change the default scope value in the Install-Package provider function as users can use the PackageManagement Install-Package cmdlet with PowerShellGet provider name.

bmanikm (changes requested):
:clock1:

Tests/PowerShellGet.Tests.ps1
alerickson: @bmanikm So what should I do to ensure both admin and non-admin tests run?  Should I only include admin tests?

bmanikm: Please take a look at https://github.com/PowerShell/PowerShellGet/blob/development/Tests/PSGetInstallModule.Tests.ps1#L522 for running the tests as non-admin.

alerickson:
@bmanikm can you review recent changes?  Core tests are temporarily skipped until dotnet in appveyor environment is updated.

PowerShellGet/private/functions/Install-PackageUtility.ps1
```diff
@@ -34,6 +34,7 @@ function Install-PackageUtility
     $Scope = $null
     $NoPathUpdate = $false
     $AcceptLicense = $false
+    $script:IsWindowsOS = (-not (Get-Variable -Name IsWindows -ErrorAction Ignore)) -or $IsWindows
```
bmanikm: >$script:IsWindowsOS = (-not (Get-Variable -Name IsWindows -ErrorAction Ignore)) -or $IsWindows [](start = 4, length = 94)

Any specific reason for not using $script:IsWindows defined in .\PowerShellGet\private\modulefile\PartOne.ps1 ?

alerickson: I made that change, please let me know if there's other changes I should make before pushing.

Tests/PSGetInstallModule.Tests.ps1
```diff
@@ -165,6 +170,201 @@ Describe PowerShell.PSGet.InstallModuleTests -Tags 'BVT','InnerLoop' {
...
+        if ($script:IsWindowsOS) {
+            $PSprocess = "PowerShell.exe";
+        }
+
+        $NonAdminConsoleOutput = Join-Path ([System.IO.Path]::GetTempPath()) 'nonadminconsole-out.txt'
+
+        Start-Process $PSprocess -ArgumentList '$null = Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force -Scope CurrentUser;
+                                                              $null = Import-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force;
+                                                              if(-not (Get-PSRepository -Name INTGallery -ErrorAction SilentlyContinue)) {
+                                                                Register-PSRepository -Name INTGallery -SourceLocation https://dtlgalleryint.cloudapp.net/api/v2/ -InstallationPolicy Trusted
```
bmanikm: >https://dtlgalleryint.cloudapp.net/api/v2/ [](start = 119, length = 42)

Please use www.poshtestgallery.com here

PowerShellGet/public/psgetfunctions/Install-Module.ps1
kilasuit: I think that whilst yes you could do that, the majority would not drop down to use Install-Package instead of Install-Module unless really needed, which I can't think of a reason off the top of my head as to why anyone would

alerickson:
@JKeithB can you review the changelog?  I included scope changes made to "install-package", but it's a provider function so I'm not sure how that should be treated.

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+## 2.0.0
+Breaking Change
+- For both administrative privileges and non-administrative privileges, default user scope for Install-Module, Install-Script, and Install-Package is now current user when running PowerShell version 6.0.0 and up.  For versions less than 6.0.0 administrative privileges will default to all users and non-administrative privileges will default to current user.
+
```
ghost: I would separate these differently. Suggested rewording accounts for the fact that most users are on PowerShell 5.1:
Default user scope for Install-Module, Install-Script, and Install-Package has changed. For Windows PowerShell (version 5.1), the default user scope is AllUsers when running in an elevated session, and CurrentUser at all other times. 
For  PowerShell version 6.0.0 and above, the default user scope is always CurrentUser.

PowerShellGet/public/psgetfunctions/Install-Script.ps1
```diff
@@ -87,19 +87,29 @@ function Install-Script
...
         Get-PSGalleryApiAvailability -Repository $Repository
 
-        if(-not (Test-RunningAsElevated) -and ($Scope -ne "CurrentUser"))
+        if($Scope -eq "AllUsers" -and -not (Test-RunningAsElevated))
         {
-            # Throw an error when Install-Script is used as a non-admin user and '-Scope CurrentUser' is not specified
-            $AdminPrivilegeErrorMessage = $LocalizedData.InstallScriptNeedsCurrentUserScopeParameterForNonAdminUser -f @($script:ProgramFilesScriptsPath, $script:MyDocumentsScriptsPath)
-            $AdminPrivilegeErrorId = 'InstallScriptNeedsCurrentUserScopeParameterForNonAdminUser'
+            # Throw an error when Install-Script is used as a non-admin user and '-Scope AllUsers'
+            $message = $LocalizedData.InstallScriptAdminPrivilegeRequiredForAllUsersScope -f @($script:programFilesModulesPath, $script:MyDocumentsModulesPath)
```
bmanikm: >@($script:programFilesModulesPath, $script:MyDocumentsModulesPath) [](start = 93, length = 66)

These vars are for modules, please use @($script:ProgramFilesScriptsPath, $script:MyDocumentsScriptsPath) here. #Closed

alerickson:
@bmanikm the tests are no longer hanging.  I believe everything should be good to go now.

bmanikm (approved):

PR #320, opened 2018-09-05
Update PowerShellGet tests to use www.poshtestgallery.com

bmanikm:


alerickson (approved):

farehar (approved):
Should we also ask/inform powershell team to change their tests.

PR #343, opened 2018-10-07
WIP: Fix Catalog tests when running with code in the cloned tree

edyoung:
I want to ensure that it works to run the tests against the code that is in the main tree (rather than only after publishing), because it makes for a quicker compile/test cycle. 

At present, when you set $env:psmodulepath to include clonepath, as suggested in the README.md, some of the catalog tests fail. This is because clonepath\tests\testmodules is now under $env:psmodulepath, and Install-Module ContosoServer fails because ContosoServer is already 'installed' (it isn't really, but Get-Module -ListAvailable picks it up).

Tried a couple of ways to fix this, but best so far is to move all the module code under src. Then $env:psmodulepath can include clonepath\src without picking up the test modules. Hence large number of files 'changed' but they have just moved directory.

Also some utility funcs which are not used yet but will hopefully be useful in future.

Thoughts?


Tests/PSGetTestUtils.psm1
```diff
@@ -1520,3 +1520,86 @@ ValidityPeriodUnits = "1"
...
+    $ErrorActionPreference="Stop"
+
+    Add-LocalTreeInPSModulePath
+
+    # Normally we want to test the code in this repo, not some other version of PowerShellGet on the system
+    $expectedModuleBase= Join-Path -Path (Split-Path $psscriptroot -parent) -ChildPath "PowerShellGet"
+
+    Write-Verbose "Ensure we load PowerShellGet from $expectedModuleBase"
+    
+    $psgetmodule = Import-Module -Name PowerShellGet -PassThru -Scope Global -Force
```
bmanikm, line 1572: >$psgetmodule = Import-Module -Name PowerShellGet -PassThru -Scope Global -Force [](start = 4, length = 79)

PackageManagement may have imported the PowerShellGet package provider from Program Files path.
Consider re-importing the PowerShellGet package provider too to ensure that it gets loaded from the clone stc path.

Tests/PSGetTestUtils.psm1
```diff
@@ -1520,3 +1520,86 @@ ValidityPeriodUnits = "1"
...
+
+function Remove-LocalTreeInPSModulePath() {
+    $modulepath= Get-LocalModulePath
+    $paths = $env:PSModulePath -split ";"
+    if ($paths[0] -like $modulepath)
+    {
+        $env:PSModulePath = ($paths[1..($paths.Length-1)] -join ";")
+    }
+    Write-Verbose "New PSModulePath: $($env:psmodulepath -replace ";",`"`n`")"
+}
```
bmanikm, line 1552: Consider handling below cases if possible:
- repo\src got added manually to the PSModulePath.
- repo\src\ with backslash got added manually.
- there are some additional module paths before repo\src in the PSModulePath.

bmanikm (approved):
:shipit:

alerickson:
Why are the files being moved into a new src directory?  Can't $env:psmodulepath include clonepath\PowerShellGet?

alerickson (approved):

Tests/PSGetTestUtils.psm1
edyoung: I'm thinking nothing too awful will happen, apart from the same dir being in psmodulepath twice...

PR #351, opened 2018-10-24
Check for clobbered commands faster

edyoung:
installing awspowershell takes 309 seconds without this change, 33 seconds with it.

Need to test some more scenarios but this looks promising. Comments?

### DIFF
--- a/src/PowerShellGet/private/functions/Validate-ModuleCommandAlreadyAvailable.ps1
+++ b/src/PowerShellGet/private/functions/Validate-ModuleCommandAlreadyAvailable.ps1
@@ -50,13 +50,14 @@ function Validate-ModuleCommandAlreadyAvailable
             # To avoid that, appending '*' at the end for each name then comparing the results.
             $CommandNames = $CurrentModuleInfo.ExportedCommands.Values.Name
 
-            # construct a regex matching any of the commands in this module.
-            $Matcher = [regex](($CommandNames | % { "($_)" }) -join "|")
+            # construct a hash with all of the commands in this module.
+            $CommandNameHash = @{}
+            $CommandNames | % { $CommandNameHash[$_] = 1 }
             
             $AvailableCommands = Microsoft.PowerShell.Core\Get-Command -Name *  `
                                                                       -ErrorAction Ignore `
                                                                       -WarningAction SilentlyContinue |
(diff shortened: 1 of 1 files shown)


src/PowerShellGet/private/functions/Validate-ModuleCommandAlreadyAvailable.ps1
```diff
@@ -49,12 +49,15 @@ function Validate-ModuleCommandAlreadyAvailable
...
             # To avoid that, appending '*' at the end for each name then comparing the results.
             $CommandNames = $CurrentModuleInfo.ExportedCommands.Values.Name
-            $CommandNamesWithWildcards = $CommandNames | Microsoft.PowerShell.Core\Foreach-Object { "$_*" }
 
-            $AvailableCommands = Microsoft.PowerShell.Core\Get-Command -Name $CommandNamesWithWildcards `
+            # construct a hash with all of the commands in this module.
+            $CommandNameHash = @{}
+            $CommandNames | % { $CommandNameHash[$_] = 1 }
+            
+            $AvailableCommands = Microsoft.PowerShell.Core\Get-Command -Name *  `
```
bmanikm: >-Name * [](start = 71, length = 7)

Consider removing '-Name *' on this to get additional perf improvement.

src/PowerShellGet/private/functions/Validate-ModuleCommandAlreadyAvailable.ps1
```diff
@@ -49,12 +49,15 @@ function Validate-ModuleCommandAlreadyAvailable
             # Get-Command loads the module if a command is already available.
             # To avoid that, appending '*' at the end for each name then comparing the results.
```
bmanikm: Consider removing this comment

bmanikm (approved):
:shipit:

alerickson (approved):

src/PowerShellGet/private/functions/Validate-ModuleCommandAlreadyAvailable.ps1
edyoung: OK. It didn't make a substantial difference to timing. What I did notice is that Get-Command -Name * returns more results than Get-Command - I don't understand why. for example get-command -name ac returns an alias for Add-content, but Get-Command | ? Name -eq 'ac' doesn't find it. Seems odd.

Issue #350, opened 2018-10-18
Register-PSRepository -Default fails silently on undocumented URL ping.


BatmanAMA:
Description of the issue
------------------
Register-PSRepository pings an undocumented endpoint(https://go.microsoft.com/fwlink/?LinkID=397631&clcid=0x409). When this ping-endpoint fails for any reason, it doesn't throw an error. 

I also would recommend that this function ping what the endpoint resolves to (https://www.powershellgallery.com/api/v2/items/psscript) instead so that firewall rules allowing https://www.powershellgallery.com/ work without needing to know this other URL. If that can't/won't be done, the additional URL should be added to the documentation for register-psrepository.

Steps to reproduce
------------------

Configure a third party firewall (we have Palo Alto, but I don't think that would matter) to drop internet based traffic without a specific matching rule. Add in a rule to allow **https** traffic to **www.powershellgallery.com**

```powershell
Register-PSRepository -Default
```

Expected behavior
-----------------

Either:
1. An Error
2. Get-PSRepository should now appear as below:
```powershell
Name                      InstallationPolicy   SourceLocation
----                      ------------------   --------------
PSGallery                 Untrusted            https://www.powershellgallery.com/api/v2
```

Actual behavior
---------------

No output to Error, Warning or Information channels from **Register-PSRepository** and afterwards, the output of **Get-PSRepository** has not changed (if it was empty, it remains so. If you had other repos, only they are shown.)

I set **$DebugPreference** to **continue** and I see it's trying to ping a different url (https://go.microsoft.com/fwlink/?LinkID=397631&clcid=0x409) entirely. This fails due to the default deny of the firewall and the command returns $Null instead of throwing an error. 

### debug output
```powershell
PS C:\Users\me> $DebugPreference = 'Continue'
PS C:\Users\me> Register-PSRepository -Default -Verbose
DEBUG: 00:00:00.0000013 Calling New() : MethodName = 'GetDynamicOptions'
DEBUG: 00:00:00.0000637 Verbose: True
DEBUG: 00:00:00.0001013 Name: PSGallery
DEBUG: 00:00:00.0063406 INVOKING PowerShell Fn Get-DynamicOptions with args Provider that has length 1
DEBUG: 00:00:00.0105389 In PowerShellGet Provider - 'Get-DynamicOptions'.
DEBUG: 00:00:00.0157741 Done calling powershell «Get-DynamicOptions» «PSModule»
[...]

Issue #317, opened 2018-08-30
Register-PSRepository raises errors about "invalid Web Uri", despite underlying httpclient recognizing the Uri and [Uri] parsing


jzabroski:
I have a very simple problem I want to solve: Due to reasons beyond my control, the company I work for uses a ".local" domain name internally, for everything. My repository url is "http://packages.mycompany.local/nuget/Mycompany-Development/api/v2" - which according to Register-PSRepository is not a "Web Uri" (whatever the heck that means).

It appears this is a bogus error message, largely due to the sophisticated and overly complicated sanity checks Register-PSRepository tries to perform.  I've extracted all the required sub-functions to replicate the behavior, modulo the final error trap that *surfaces the wrong error*.

```powershell
function Get-LocationString
{
    [CmdletBinding(PositionalBinding=$false)]
    Param
    (
        [Parameter()]
        [Uri]
        $LocationUri
    )

    $LocationString = $null

    if($LocationUri)
    {
        if($LocationUri.Scheme -eq 'file')
        {
            $LocationString = $LocationUri.OriginalString
        }
        elseif($LocationUri.AbsoluteUri)
        {
            $LocationString = $LocationUri.AbsoluteUri
        }
        else
        {
            $LocationString = $LocationUri.ToString()
        }
    }

    return $LocationString
}

function Resolve-Location
{
    [CmdletBinding()]
    [OutputType([string])]
    Param
    (
        [Parameter(Mandatory=$true)]
        [string]
        $Location,

        [Parameter(Mandatory=$true)]
        [string]
        $LocationParameterName,

        [Parameter()]
        $Credential,

        [Parameter()]
        $Proxy,

        [Parameter()]
        $ProxyCredential,

        [Parameter()]
        [System.Management.Automation.PSCmdlet]
        $CallerPSCmdlet
    )

    # Ping and resolve the specified location
    if(-not (Test-WebUri -uri $Location))
    {
        if(Microsoft.PowerShell.Management\Test-Path -Path $Location)
        {
            return $Location
        }
        elseif($CallerPSCmdlet)
        {
            $message = $LocalizedData.PathNotFound -f ($Location)
            ThrowError -ExceptionName "System.ArgumentException" `
                       -ExceptionMessage $message `
                       -ErrorId "PathNotFound" `
                       -CallerPSCmdlet $CallerPSCmdlet `
                       -ErrorCategory InvalidArgument `
                       -ExceptionObject $Location
        }
    }
    else
    {
[...]

PR #352, opened 2018-10-26
Warn when a repo is unreachable, don't produce a confusing error

edyoung:
Partially addresses #317 by improving the message, and producing a non-blocking warning rather than an error


edyoung:
From the CI output, looks like I broke something. Will need to investigate further before merging anything.

edyoung:
Also resolves #350. I still need to look at wmf 4 failures, but let me know what you think @bmanikm @alerickson

bmanikm (approved):
:shipit:

alerickson (approved):

PR #358, opened 2018-11-12
Script source/publish location fix

alerickson:


tools/build.psm1
```diff
@@ -208,7 +208,7 @@ function Invoke-PowerShellGetTest {
 
     $PesterTag = '' # Conveys all test priorities
     if (-not $IsFullTestPass) {
-        $PesterTag = 'BVT' # Only BVTs
+        #$PesterTag = 'BVT' # Only BVTs
```
edyoung: Did you mean to check in this change?

edyoung (approved):
Thanks!

tools/build.psm1
alerickson: Yes, I just want to make sure all the tests (the daily tests) pass.  I'll uncomment with the last commit.

src/PowerShellGet/private/functions/Get-PublishLocation.ps1
```diff
@@ -8,34 +8,9 @@ function Get-PublishLocation
...
-            {
-                $tempPublishLocation = $Location + 'package/'
-            }
-
-            if($tempPublishLocation)
-            {
-                $PublishLocation = $tempPublishLocation
-            }
-        }
+        return $Location
```
bmanikm, line 14: >return $Location [](start = 8, length = 16)

Consider specifying this as a breaking change in the release notes as this breaks the  existing automation scripts for registering the repositories.

PR #359, opened 2018-11-13
Add tab completion for Publish-Module -Name

matt9ucci:
The completion values come from `Get-Module -ListAvailable`.

I did not add a new function like [`Add-ArgumentCompleter`](src/PowerShellGet/private/functions/Add-ArgumentCompleter.ps1). Because the completer is just for `Publish-Module -Name`, not for the other *-Module functions (Find/Install/Save-Module).


bmanikm (approved):
Thanks!

PR #370, opened 2018-11-21
Use TestDrive in Pester.PSGetModuleHelp.Tests.ps1

matt9ucci:
As written in the PowerShell's [testing guideline](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#test-drive), it is preferred to use Pester's TestDrive to custom temp dir.


edyoung:
This change looks fine - thanks! But we have some other failing tests which are causing CI build to fail. Will merge this once that is resolved.

BTW It would not be quite as straightforward to use TestDrive: for all of the places where PowerShellGet uses temp files, because the package provider code runs in a separate runspace created by package management module, and that code can't access the testdrive variable.

PR #371, opened 2018-11-21
some extra checks to debug CI failure

edyoung:


edyoung:
We determined that the failing tests are due to an issue in packagemanagement

PR #386, opened 2018-12-18
 fix New-scriptfileinfo command

whiggs:
This pull request fixes the new-scriptfileinfo command in the powershellget powershell module.  Currently, when the aforementioned command is run, the file it generates contains invalid syntax that causes the "Test-scriptfileinfo" command to fail when the generated file is passed to it as a parameter:
image

However, the changes I have made generate script file info which can be published to the psgallery:
image

It is important to note that "test-scriptfileinfo" does not correctly parse files generated by "new-scriptfileinfo", as files generated with "new-scriptfileinfo" are successfully passed to "test-scriptfileinfo", but fail to publish with error.
image

### DIFF
--- a/src/PowerShellGet/private/functions/Get-ScriptCommentHelpInfoString.ps1
+++ b/src/PowerShellGet/private/functions/Get-ScriptCommentHelpInfoString.ps1
@@ -50,7 +50,7 @@ function Get-ScriptCommentHelpInfoString {
 
     Process {
         If ($newscriptnfo) {
-            $ScriptCommentHelpInfoString = "`r`n`r`n.DESCRIPTION `r`n $Description `r`n`r`n"
+            $ScriptCommentHelpInfoString = ".DESCRIPTION `r`n $Description `r`n`r`n"
         }
         else {
             $ScriptCommentHelpInfoString = "<# `r`n`r`n.DESCRIPTION `r`n $Description `r`n`r`n"
@@ -108,13 +108,7 @@ function Get-ScriptCommentHelpInfoString {
             $ScriptCommentHelpInfoString += ".FUNCTIONALITY `r`n$($Functionality -join "`r`n") `r`n`r`n"
         }
 
-        If ($newscriptnfo) {
-            $ScriptCommentHelpInfoString += "`r`n"
-        }
-        else {
-            $ScriptCommentHelpInfoString += "#> `r`n"
-        }
-
+        $ScriptCommentHelpInfoString += "#> `r`n"
         return $ScriptCommentHelpInfoString
     }
 }
--- a/src/PowerShellGet/private/functions/Get-ScriptCommentHelpInfoString.ps1
+++ b/src/PowerShellGet/private/functions/Get-ScriptCommentHelpInfoString.ps1
@@ -1,9 +1,8 @@
-function Get-ScriptCommentHelpInfoString
-{
-    [CmdletBinding(PositionalBinding=$false)]
+function Get-ScriptCommentHelpInfoString {
+    [CmdletBinding(PositionalBinding = $false)]
     Param
     (
-        [Parameter(Mandatory=$true)]
+        [Parameter(Mandatory = $true)]
         [ValidateNotNullOrEmpty()]
         [string]
         $Description,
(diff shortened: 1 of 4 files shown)


whiggs:
And while the appveyor and travis tests failed for some reason, I can assure you that, when publishing and actual script to an actual repository, it works:
https://vimeo.com/user66875139/review/307087502/97e8ec3b73

alerickson:
@whiggs the changes are not merged because they are breaking tests.  We need to resolve those issues first.  Please feel free to make the necessary changes, otherwise I can take a look when I have a moment.

whiggs:
@alerickson You do realize that the fact that the module was "passing" tests when the "test-scriptfileinfo" and "new-scriptfileinfo" tests were completely, 100% broken, right?

whiggs:
Furthermore, I am fairly certain that the reason that the tests are "failing" is because the latest version of the packagemanagement module is not being used with powershellget.  Take a look at the first errors that are being generated in travis:
image

And compare that to the first errors I get when running that particular task locally:
image

In short, I don't think the build server is running the tests nearly as well as you think they do

whiggs:
And thus, my major problem with only merging code based on the results of these "tests":
1. Code that does not work in production is able to pass these "tests"
2. I have, in my above posts, provided you with all of the information you could possibly need to:
        1.  identify what the issue is
        2.  Show how the issue prevents the module from working as expected
        3.  Show the changes that I made to the code
        4.  See how the changes that I made fix the aforementioned issue, where previously it did NOT 
             work.  
3.  They are not effective in determining if submitted code is actually functional, but rather, promotes the submission of code that is broken over code that works.
4.  Delays, if not altogether prevents, the release of a functional product, which is a disservice to all who rely on its functioning properly.
5.  Is a disservice, deterrent, and an insult quite frankly for developers who try to HELP YOU fix your product.

How about, rather than relying on the results of these "tests" to determine if code works or not 1) confirm the issue is present for yourselves (on your machines), and, once the issue is confirmed, download the module I submitted, load it into the powershell session, and try again and see if it works.  I think you will find the results that you get will be very similar to the results I displayed in previous posts (check out the link to vimeo).

edyoung:
William, thanks for your contribution.

I'm out of the office so I haven't had a chance to look at this change yet, but I want to try to clarify the process. We have a regression test suite, and we try to keep it passing all the time. This helps to make sure we don't accidentally break something. 
So when tests fail with a new change, we need to investigate, determine the root problem, and fix it as part of the same merge. It's not an insult, this is a standard practice. The same process applies whoever submits the change. 

The tests are code, and code can have bugs, so it's very possible that in this case the tests are incorrect. We don't blindly trust them, but manually investigating the issue (as you suggest above) is obviously a little bit more time-consuming than just reviewing the code. We'll look at it, but it may take longer to merge your PR than it would otherwise.

src/PowerShellGet/private/functions/Get-ScriptCommentHelpInfoString.ps1
```diff
@@ -42,80 +41,74 @@ function Get-ScriptCommentHelpInfoString
 
         [Parameter()]
         [string]
-        $Functionality
+        $Functionality,
+
+        [Parameter()]
+        [Switch]
+        $newscriptnfo
```
edyoung, line 48: Typo in parameter name. NewScriptInfo would be better; or it would be more descriptive called $SkipSurroundingComment or similar

src/PowerShellGet/public/psgetfunctions/Test-ScriptFileInfo.ps1
```diff
@@ -1,120 +1,122 @@
...
-                        -ErrorCategory InvalidArgument
+                -ExceptionMessage $errorMessage `
+                -ErrorId "InvalidScriptFilePath" `
+                -CallerPSCmdlet $PSCmdlet `
+                -ExceptionObject $scriptFilePath `
+                -ErrorCategory InvalidArgument
             return
         }
 
+        Write-Host "start psscript"
```
edyoung, line 68: Need to remove this debugging output, or convert to Write-Debug, same for other instances in this file

src/PowerShellGet/public/psgetfunctions/Test-ScriptFileInfo.ps1
```diff
@@ -1,120 +1,122 @@
-function Test-ScriptFileInfo
-{
+function Test-ScriptFileInfo {
```
edyoung, line 1: Your updated Test-ScriptFileInfo reports that previously-valid scripts are invalid, which is why tests are failing. To see this, save https://www.poshtestgallery.com/packages/Fabrikam-ServerScript/2.5/Content/Fabrikam-ServerScript.ps1 to a location on your disk, and run your version of Test-ScriptFileInfo against it: you'll see an error, whereas the shipping version of Test-ScriptFileInfo does not complain. 

Your updated structure where .DESCRIPTION is in the same comment block makes more sense to me than having 2 comments, *but* we can't afford to break existing scripts which users have already shipped, so we can't merge this change unless it continues to parse scripts as they currently exist in the wild. 

To keep back-compat, we need things to work in 2 ways:
1) New scripts created with your updated New-ScriptFileInfo need to be OK when tested with an existing Test-ScriptFileInfo

2) Existing scripts created with old New-ScriptFileInfo need to be OK when tested with the new Test-ScriptFileInfo

This is necessary because we can't force everyone to upgrade simultaneously.

edyoung:
Closing this as inactive - can't merge as-is due to open issues above. Feel free to reactivate, especially if we can clarify exactly what the problem is

whiggs:
screw you guys.

edyoung:
I have explained why we can't merge the change as it is. Not trying to be difficult, we just don't want to break back-compat. If you disagree, reopen and explain why.

PR #395, opened 2019-01-10
Don't publish hidden files (eg .git dir)

### LINKED ISSUE #378: .git folders being included in some uploaded modules
Many packages (https://www.powershellgallery.com/packages/DosInstallUtilities.Menu/2.15 for example) contain .git folders. Looks like these are not being filtered out by the dotnet or nuget commands being used for publishing. We could copy all the files to be published to a new temp dir and leave out all the files which should be ignored (files starting with . or with hidden attribute on windows).

edyoung:
Resolves #378


edyoung:
oops, why are all these other commits in here? I will tidy up and try again

Issue #394, opened 2019-01-07
Update-ModuleManifest complains when adding an invalid entry value, yet still updates the manifest.


pougetat:
This is a regression from Windows PowerShell.

This has only been tried with one entry, RootModule, where the bug only surfaces if the input manifest file does not have the entry being updated yet.

Steps to reproduce
------------------

```powershell
# Important: create the manifest *without* -RootModule
New-ModuleManifest tmp.psd1
# Try to update (add) the RootModule entry with an invalid value.
Update-ModuleManifest tmp.psd1 -RootModule '\/'
'---'
# Extract the now-current RootModule entry.
Select-String ^RootModule tmp.psd1
```

Expected behavior
-----------------

```
Microsoft.PowerShell.Core\Test-ModuleManifest : The module manifest '\/' could not be processed because it is not a valid PowerShell module manifest file.
...
---
```

Actual behavior
---------------

```
Microsoft.PowerShell.Core\Test-ModuleManifest : The module manifest '\/' could not be processed because it is not a valid PowerShell module manifest file.
...
---

tmp.psd1:12:RootModule = '\/'
```
That is, the manifest file was unexpectedly still updated with the invalid entry.

Environment data
----------------
```
PowerShell Core 6.2.0-preview.1 on macOS 10.14.1
PowerShell Core 6.2.0-preview.1 on Ubuntu 16.04.5 LTS
PowerShell Core 6.2.0-preview.1 on Microsoft Windows 10 Pro (64-bit; Version 1803, OS Build: 17134.345)
```

pougetat:
@SteveL-MSFT Hey is this issue up for grabs ? If it is I was wondering if I could work on it :)

SydneyhSmith:
@pougetat we would love it if you worked on this issue! Feel free to assign it to yourself.

pougetat:
Hey,

I copied this issue from the PowerShell repo and cannot actually reproduce it using just PowerShellGet. The issue is indeed from PowerShell and not this repo so I am going to go ahead and close this issue.

See https://github.com/PowerShell/PowerShell/issues/8389

:)

alerickson:
Thanks @pougetat!

PR #398, opened 2019-01-11
Throw error when Test-ModuleManifest fails in Update-ModuleManifest

pougetat:
Fix for #394


edyoung:
looks good to me, thanks

Issue #402, opened 2019-01-16
Update-ModuleManifest prepends 'PSGet_' to the module name on line 2


ThePoShWolf:
On an existing module manifest line 2 may look like this:

```
# Module manifest for module 'ModuleName'
```

After running Update-ModuleManifest <path to manifest>, line two will appear like this:

```
# Module manifest for module 'PSGet_ModuleName
```

This appears to be a result of prepending 'PSGet_' to the module manifest in $tempPath on line 677.

Steps to reproduce
------------------

```powershell
New-ModuleManifest .\Test.psd1
Update-ModuleManifest .\Test.psd1
```

Expected behavior
-----------------

```none
Line 2:
# Module manifest for module 'Test'
```

Actual behavior
---------------

```none
Line 2:
# Module manifest for module 'PSGet_Test'
```

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      5.1.17134.407
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17134.407
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

```

```powershell
> Get-Module -ListAvailable PowerShellGet,PackageManagement
ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.2.4      PackageManagement                   {Find-Package, Get-Package, Get-PackageProvider, Get-PackageSource...}
Script     1.2.2      PackageManagement                   {Find-Package, Get-Package, Get-PackageProvider, Get-PackageSource...}
Binary     1.0.0.1    PackageManagement                   {Find-Package, Get-Package, Get-PackageProvider, Get-PackageSource...}
Script     2.0.4      PowerShellGet                       {Find-Command, Find-DSCResource, Find-Module, Find-RoleCapability...}
Script     1.0.0.1    PowerShellGet                       {Install-Module, Find-Module, Save-Module, Update-Module...}
```

```powershell
> Get-PackageProvider -ListAvailable
Name                     Version          DynamicOptions
----                     -------          --------------
msi                      3.0.0.0          AdditionalArguments
msu                      3.0.0.0
nuget                    2.8.5.208
[...]

PR #403, opened 2019-01-18
Now removes the PSGet_ prepended to line 2 in the updated manifest.

ThePoShWolf:
This resolves issue #402. A simple regex replace to remove PSGet_ in the module name on line 2.


ThePoShWolf:
Hrm. I'm unfortunately not very familiar with Pester, though it appears that this isn't related to the issue I submitted. If there is something that I can do to help troubleshoot the failing tests, let me know.

alerickson:
Thanks for the contribution!  This looks fine, there was a failure setting up a describe block, so it's not necessarily even a test failure.  I'm re-running tests now.

Issue #401, opened 2019-01-16
Set-PSRepository removes trailing / on SourceLocation 


SydneyhSmith:
Opening from twitter https://twitter.com/PrzemyslawKlys/status/1085647512508059649
Setting the PS Gallery to "Trusted" or "Default" removes the trailing \ on source location causing the repository to not be found at Install-Package

alexandair:
Causing the repository to be not be found at Update-Module.

edyoung:
I think the issue is subtly different, but it's possible I am missing something. 
```powershell
Set-PSRepository PSGallery -InstallationPolicy Trusted 
```

resets the SourceLocation to the built-in value, which is https://www.powershellgallery.com/api/v2 (no trailing /). But Register-PSRepository -Default also sets the value to https://www.powershellgallery.com/api/v2 . The question is what previously set the repository url to https://www.powershellgallery.com/api/v2/ (with a trailing /). See also #349

alexandair:
If the URL without a trailing / is the built-in value, why does Update-Module ask for an URL with the trailing /?

```powershell
PS C:\> update-module powershellget -verbose
VERBOSE: Checking for updates for module 'PowerShellGet'.
PackageManagement\Install-Package : Unable to find repository 'https://www.powershellgallery.com/api/v2/'.
Use Get-PSRepository to see all available repositories.
```

edyoung:
The url which was used at initial install time is saved, and then used again at update time. If no repository matching the saved url is found at that time, you see this error.

edyoung:
See comments on PR above. This does NOT change the behavior of Set-PSRepository but I believe it resolves the Update-Module problem.

PR #407, opened 2019-01-24
Edyoung/issue 349 catastrophic

### LINKED ISSUE #349: Catastrophic terminating error for PSGallery-installed "Update-Module" due to repo URL syntax change
# ISSUE

`SourceLocation` PSGallery default URL at some point has changed slightly in it's stored syntax.

I have about half of my systems (many) (all Win7x64 PoSH 5.1 latest) that I have been testing today that all have this as the default URL (you cannot even manually force the PSGallery repo URL):

**_(NOTE THE TRAILING SLASH ON "SourceLocation"):_**
`λ  Get-PSRepository psgallery | fl

Name                      : PSGallery
SourceLocation            : https://www.powershellgallery.com/api/v2/
Trusted                   : True
Registered                : True
InstallationPolicy        : Trusted
PackageManagementProvider : NuGet
PublishLocation           : https://www.powershellgallery.com/api/v2/package/
ScriptSourceLocation      : https://www.powershellgallery.com/api/v2/items/psscript
ScriptPublishLocation     : https://www.powershellgallery.com/api/v2/package/
ProviderOptions           : {}
`

And the other half (also MANY) with this as the default URL:

**_(NOTE NO TRAILING SLASH ON "SourceLocation"):_**
`λ  Get-PSRepository psgallery | fl

Name                      : PSGallery
SourceLocation            : https://www.powershellgallery.com/api/v2
Trusted                   : True
Registered                : True
InstallationPolicy        : Trusted
PackageManagementProvider : NuGet
PublishLocation           : https://www.powershellgallery.com/api/v2/package/
ScriptSourceLocation      : https://www.powershellgallery.com/api/v2/items/psscript
ScriptPublishLocation     : https://www.powershellgallery.com/api/v2/package/
ProviderOptions           : {}

`

Why is this relevant?  Diff the `PSGetModuleInfo.xml` from either "flavor" from each and every PSGallery-installed module, and look at line 143 and it becomes clear that this syntax difference above actually made it into the stored config files:

snag_10-13-2018_16-40-05

And why does that matter?  **Purely as an example**, if you simply copy (or distribute) all the modules installed on an Internet-facing machine to one that is not (and has the opposite URL config), and then attach it to the Internet and run a simple `Update-Module * -Verbose`, you get a xmas tree of red and catastrophic failures as shown here:

snag_10-13-2018_17-01-42

Versus this, from the half of systems with the other "flavor" PSGallery URL:

snag_10-13-2018_17-01-22

# Cause

Given the two above runs are the exact same copied PSGallery installed modules placed in `C:\Program Files\WindowsPowershell\Modules`, and both were reset/defaulted PSGallery sources as shown above, other than this needing to be addressed is when you perform a `Register-PSRepository -Default` clearly something changed along the way and I cannot find the root source of that URL.  Where is it defined so that I can even look at these systems to determine if the URL definition is different or whether it was a specific code/version issue?

The catastrophic issue risk is if later it is ever necessary to re-register the PSGallery using the command `Register-PSRepository -Default` after unregistering.  Because, when the URL then is changed to the (apparently) newer no trailing slash, 100% of the already installed PSGallery modules will then throw the above terminating errors upon attempting to update.  At that point, the only workaround I have found is to first have to determine each and every one who's PSGetModuleInfo.xml is now incorrect and was stored with the trailing slash, and then force a fresh re-install and not update using `Install-Module -Force`.  At that point, I can confirm the new `PSGetModuleInfo.xml` is stored with whatever is the current repo URL, so without the slash.  Then and only then will `Update-Module` ever work on that system again for those modules.

To further prove this root cause, one can also manually edit the `PSGetModuleInfo.xml` and remove the erroneous legacy trailing slash from the PSGallery URL and without even a session restart re-run the `Update-Module * -Verbose` and there will now be no errors and all module updates will be successful.

Steps to reproduce
------------------

```powershell
Update-Module * -Verbose

VERBOSE: Checking for updates for module 'WinFormPS'.
PackageManagement\Install-Package : Unable to find repository
'https://www.powershellgallery.com/api/v2/'. Use Get-PSRepository to see all available
repositories.
At C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\2.0.0\PSModule.psm1:12979
char:20
+ ...           $sid = PackageManagement\Install-Package @PSBoundParameters
+                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Microsoft.Power....InstallPackage:InstallPa
   ckage) [Install-Package], Exception
    + FullyQualifiedErrorId : SourceNotFound,Microsoft.PowerShell.PackageManagement.Cmdlet
   s.InstallPackage

```

Expected behavior
-----------------

```none
Updating of all modules occurs for same set of PSGallery installed modules successful and fails on others.  Expect consistent gallery source URL across all versions of PoSH and module.
```

Actual behavior
---------------

```none
See Above
```

Environment data
----------------

```powershell
> $PSVersionTable

λ  $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.14409.1012
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14409.1012
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

```powershell
> Get-Module -ListAvailable PowerShellGet,PackageManagement

```
I literally deleted all versions of powershellget and packagemanagement back to 1.0.0.1 2017 (NOTE I started testing with the latest v2.0.1 with same results and worked backwards) and can not replicate a reset of the psgallery URL to now place the trailing slash on it as is shown in probably half (many) of the systems I have seen.  All WIN7x64 in this issue/testing PS5.1.  Some probably were upgraded through v4 PoSH while perhaps others went directly to 5.x but because I have not yet confirmed what is responsible for setting up the default PSGallery URL, I cannot confirm whether there is a correlation between the PoSH upgrade paths vs just being a module/assembly issue, or perhaps arbitrary code that tacked it on before storing it in the xml repo file.  However, at some point there was a combination of PoSH versions, Packagemanagement module, and Powershellget module that was allowing the trailing slash systems to update.

```powershell
> Get-PackageProvider -ListAvailable



    Directory: C:\Program Files\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.1.1.0    PackageManagement                   {Find-Package, Get-Package, Ge...
Binary     1.0.0.1    PackageManagement                   {Find-Package, Get-Package, Ge...
Script     1.0.0.1    PowerShellGet                       {Install-Module, Find-Module, ...

```

edyoung:
Should resolve #349 and #401 

When a module is installed, we store the SourceLocation at that time. When you run Update-Module, it looks for a source which matches that saved location. When the two do not match, the update fails. The expected URL for the PowerShellGallery is https://www.powershellgallery.com/api/v2 , but some operation (it is still not clear exactly what) has led to some users having the gallery URL as https://www.powershellgallery.com/api/v2/ - with a trailing /. 

This is treated as the same URL by the web server but is different-enough that Update-Module sees the saved and current repository sources as different.

This change makes it so that when we look for a repository by URL/path we ignore trailing / or \ characters when comparing them. This should resolve the practical issues people are hitting with Update-Module.

### DIFF
--- a/Tests/PSGetUpdateModule.Tests.ps1
+++ b/Tests/PSGetUpdateModule.Tests.ps1
@@ -91,19 +91,19 @@ Describe UpdateModuleFromAlternateRepo -Tags 'BVT' {
         PSGetTestUtils\Uninstall-Module ContosoClient
     }
 
-    It "Check situation" {
+    It "Check that removing a slash from a repo doesn't break update" {
         $withSlash = "https://www.poshtestgallery.com/api/v2/"
         $noSlash = "https://www.poshtestgallery.com/api/v2"
-        Write-Host (Get-PSRepository | Out-String)
+        #Write-Host (Get-PSRepository | Out-String)
         (Get-PSRepository PSGallery).SourceLocation | Should Be $withSlash
 
         Install-Module ContosoServer -RequiredVersion 1.0
         (Get-InstalledModule ContosoServer).RepositorySourceLocation | Should Be $withSlash
-        Write-Host (Get-InstalledModule ContosoServer -AllVersions | Format-List | Out-String)
+        #Write-Host (Get-InstalledModule ContosoServer -AllVersions | Format-List | Out-String)
 
         # now update where PSGallery Source Location is
         Set-PSGallerySourceLocation -Location $noSlash
-        Write-Host (Get-PSRepository | Out-String)
+        #Write-Host (Get-PSRepository | Out-String)
         (Get-PSRepository PSGallery).SourceLocation | Should Be $noSlash
 
         # reload powershellget to force-update cached repository info
@@ -111,11 +111,39 @@ Describe UpdateModuleFromAlternateRepo -Tags 'BVT' {
 
         # now try and update module isntalled using other SourceLocation
         Update-Module ContosoServer -RequiredVersion 2.0 -ErrorAction Stop
-        Write-Host (Get-InstalledModule ContosoServer -AllVersions | Format-List | Out-String)
+        #Write-Host (Get-InstalledModule ContosoServer -AllVersions | Format-List | Out-String)
         (Get-InstalledModule ContosoServer).RepositorySourceLocation | Should Be $noSlash
 
         (Get-InstalledModule ContosoServer).Version | Should Be 2.0
+    }
 
+    It "Check that adding a slash to a repo doesn't break update" {
+        $withSlash = "https://www.poshtestgallery.com/api/v2/"
+        $noSlash = "https://www.poshtestgallery.com/api/v2"
+        #Write-Host (Get-PSRepository | Out-String)
+
+        Set-PSGallerySourceLocation -Location $noSlash
+
+        (Get-PSRepository PSGallery).SourceLocation | Should Be $noSlash
+
+        Install-Module ContosoServer -RequiredVersion 1.0
+        (Get-InstalledModule ContosoServer).RepositorySourceLocation | Should Be $noSlash
+        #Write-Host (Get-InstalledModule ContosoServer -AllVersions | Format-List | Out-String)
+
+        # now update where PSGallery Source Location is
+        Set-PSGallerySourceLocation -Location $withSlash
+        #Write-Host (Get-PSRepository | Out-String)
+        (Get-PSRepository PSGallery).SourceLocation | Should Be $withSlash
+
+        # reload powershellget to force-update cached repository info
+        Import-Module PowerShellGet -Force
+
+        # now try and update module isntalled using other SourceLocation
+        Update-Module ContosoServer -RequiredVersion 2.0 -ErrorAction Stop
+        #Write-Host (Get-InstalledModule ContosoServer -AllVersions | Format-List | Out-String)
+        (Get-InstalledModule ContosoServer).RepositorySourceLocation | Should Be $withSlash
+
+        (Get-InstalledModule ContosoServer).Version | Should Be 2.0
     }
 }
 
--- a/Tests/PSGetUpdateModule.Tests.ps1
+++ b/Tests/PSGetUpdateModule.Tests.ps1
@@ -77,6 +77,48 @@ function SuiteCleanup {
     }
 }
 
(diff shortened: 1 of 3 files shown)


farehar (approved):

sg1970:
Can someone please let me know if this issue is fixed or not ? If fixed where/what is the fix ? From reading the associated thread it appears that the issue is due to the slash at the end of the URL ( https://www.powershellgallery.com/api/v2 )  but whether I have a slash or not it doesn't seem to make any difference for me and I get the same error.

edyoung:
A fix exists but a binary package containing that fix has not yet been released - hopefully very soon now.

sg1970:
Thanks

sg1970:
> A fix exists but a binary package containing that fix has not yet been released - hopefully very soon now.

Any word on this yet ?

edyoung:
Use PowerShellGet 2.1.0 or above, out now wherever PowerShell modules are sold ;-)

sg1970:
Thanks it works !

Dyermull:
> Use PowerShellGet 2.1.0 or above, out now wherever PowerShell modules are sold ;-)

Respectfully, how do I update PowerShellGet to the latest version when the Install-Module functionality is broken and running Update-Module -Name PowerShellGet produces an error "Update-Module : Module 'powershellget' was not installed by using Install-Module, so it cannot be updated." (Server 2012 R2 shipped with WMF 4.0 and I upgraded to WMF 5.1 a while back).

It kind of feels like a catch-22. I've been struggling through issues with PowerShellGet's Install-Module for a couple of weeks until I came across this thread and it's relatives, but still haven't been able to fix it. Any advice in getting PowerShellGet updated in my situation would be much appreciated.

alerickson:
@Dyermull Can you please file an issue for this with repro steps?  I'll take a look at this

Dyermull:
@alerickson Will do, thanks!

Issue #409, opened 2019-01-25
Update-ModuleManifest -CmdletsToExport, -AliasesToExport, and -FunctionsToExport empty array not allowed


natescherer:
Is there a design reason that you can't provide an empty array to -CmdletsToExport and -AliasesToExport for Update-ModuleManifest?

Best practice for these two values in the manifest is to set an empty array for these if there aren't any to export, but it doesn't seem possible to do via Update-ModuleManifest.

Looking in the code I see both parameters have [ValidateNotNullOrEmpty()] set.

If there isn't a design reason for this, I'll fork and see if I can get this working.

Thanks!

natescherer:
This also applies to -FunctionsToExport now that I think about it. Title updated.

alerickson:
Don't believe there's a design reason for this, feel free to work on it.

natescherer:
@pougetat Beat me to the punch on this, looks like. Closing.

PR #414, opened 2019-02-01
Update-ModuleManifest can take an empty array of functions or cmdlets to export #409

pougetat:
Fix for #409


alerickson:
Thanks @pougetat!

Issue #373, opened 2018-11-25
NestedModules array is cleared after calling Update-ModuleManifest


lpatalas:
Hello.

I don't know if it's a bug or I just don't understand this behavior but it seems unexpected.

Steps to reproduce
------------------

https://gist.github.com/lpatalas/914a9fcafa6810cf369b118f3533ac67

```powershell
$originalManifestPath = Join-Path $PSScriptRoot 'Original.psd1'
$updatedManifestPath = Join-Path $PSScriptRoot 'Updated.psd1'

'function Get-Root { Write-Host "Root" }' | Set-Content -Path 'RootModule.psm1'
'function Get-Nested { Write-Host "Nested" }' | Set-Content -Path 'NestedModule.psm1'

New-ModuleManifest `
    -Path $originalManifestPath `
    -NestedModules @('NestedModule.psm1') `
    -RootModule 'RootModule.psm1' `
    -FunctionsToExport @('Get-Root', 'Get-Nested') `
    -CmdletsToExport @() `
    -VariablesToExport @() `
    -AliasesToExport @()

Copy-Item `
    -Path $originalManifestPath `
    -Destination $updatedManifestPath `
    -Force

Update-ModuleManifest `
    -Path $updatedManifestPath

$originalManifest = Import-PowerShellDataFile -Path $originalManifestPath
$updatedManifest = Import-PowerShellDataFile -Path $updatedManifestPath

Write-Host 'Original:' $originalManifest.NestedModules
Write-Host 'Updated:' $updatedManifest.NestedModules
```

Expected behavior
-----------------

`NestedModules` value should not change after update.

```powershell
# Modules to import as nested modules of the module specified in RootModule/ModuleToProcess
NestedModules = @('NestedModule.psm1')
```

Actual behavior
---------------

`NestedModules` are set to empty array and commented-out.

```powershell
# Modules to import as nested modules of the module specified in RootModule/ModuleToProcess
# NestedModules = @()
```

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.1.1
PSEdition                      Core
GitCommitId                    6.1.1
OS                             Microsoft Windows 10.0.17134
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

```powershell
[...]

PR #415, opened 2019-02-02
Fix Update-ModuleManifest clears FunctionsToExport, AliasesToExport, NestedModules #373

pougetat:
This is a fix for #373 .

Update : I've written a comment on the issue which essentially say that this was resolved in a recent release of PowerShell. Should I close this PR ?

### DIFF
--- a/Tests/PSGetUpdateModuleManifest.Tests.ps1
+++ b/Tests/PSGetUpdateModuleManifest.Tests.ps1
@@ -155,7 +155,7 @@ Describe PowerShell.PSGet.UpdateModuleManifest -Tags 'BVT','InnerLoop' {
     #
     It UpdateModuleManifestWithNoAdditionalParameters3 {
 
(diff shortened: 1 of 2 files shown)


pougetat:
I'm still working on fixing the test.
Having said that here is the explanation for why this issue exists : 
- Update-ModuleManifest calls Test-ModuleManifest to get a representation of the original manifest file created.
- For some reason Test-ModuleManifest returns no NestedModules even though the NestedModule given in the issue description seems to be totally valid.
- As a result Update-ModuleManifest sets the manifest file's nested modules field to null.

alerickson:
For the failing test, try using "Microsoft.PowerShell.Management" or "Microsoft.PowerShell.Utility", since they're nested modules used elsewhere in the file.   I think throwing an error makes sense when a nested module's new manifest returns as invalid.  Clearing nested modules after that sounds reasonable, though.  As @edyoung commented in the related issue, would still like to merge this PR for older versions of PS.

pougetat:
@alerickson Sounds good. I'll update my PR to take this into account today. :)

PR #421, opened 2019-02-11
Fix #78

### LINKED ISSUE #78: Update-Module doesn't check all modules when not elevated
When a user runs `Update-Module` from a non-elevated host, it only checks modules in the `-Scope CurrentUser` folder (i.e. in your `Documents\WindowsPowerShell\Modules`, not the `Program Files` folder).

This means that there's no built-in way for a (non-elevated) user to even check that **all** of their modules are up to date.

## Expected Behavior

Even when running in a **non-elevated** host, `Update-Module` should check *all* modules for updates, even though it cannot install them:

```posh
> Get-InstalledModule | ft Name, InstalledLocation

Name               InstalledLocation
----               -----------------
PackageManagement  C:\Program Files\WindowsPowerShell\Modules\PackageManagement\1.1.1.0
Pester             C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.3
PowerShellGet      C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\1.1.1.0
PSScriptAnalyzer   C:\Users\Joel\OneDrive\Documents\WindowsPowerShell\Modules\PSScriptAnalyzer\1.7.0

> Update-Module -WhatIf
What if: Performing the operation "Update-Module" on target "Version '1.1.1.0' of module 'PowerShellGet', updating to version '1.1.2.0'".
What if: Performing the operation "Update-Module" on target "Version '1.7.0' of module 'PSScriptAnalyzer', updating to version '1.9.0'".

> Update-Module
WARNING: Skipped performing Update-Module on target "Version '1.1.1.0' of module 'PowerShellGet', to version '1.1.2.0'" because it's installed in the LocalMachine scope, which requires elevation.
```

## Current Behavior

Currently, PowerShellGet doesn't even let you know that it's not bothering to check:

```posh
> Update-Module -WhatIf
What if: Performing the operation "Update-Module" on target "Version '1.7.0' of module 'PSScriptAnalyzer', updating to version '1.9.0'".

> foreach($module in Get-InstalledModule) {
    Find-Module -Name $module.Name -Repository $module.Repository -MinimumVersion $module.Version | Where Version -ne $module.Version
}

Version  Name               Repository  Description
-------  ----               ----------  -----------
1.9.0    PSScriptAnalyzer   PSGallery   PackageManagement (a.k.a. OneGet) is a new way to discover and install software packages from around the w...
1.1.2.0  PowerShellGet      PSGallery   PowerShell module with commands for discovering, installing, updating and publishing the PowerShell artifa...
```

### In an elevated host:

1. Use Install-Module to install a specific (older) version of a module to `-Scope CurrentUser` 
2. Use Install-Module to install a specific (older) version of a module to `-Scope AllUser`
3. Use `Update-Module -Whatif` to verify that it detects both of them are out of date.

### In a non-elevated host:
4. Use `Update-Module -Whatif` to verify that it only detects one of them.
5. Use `Update-Module` to verify that it really updates only that one module.
6. Use `Update-Module -Whatif` to see that it looks like there are no modules in need of updates. 
7. Use the following script to see that the other modules **does** need updates:
```posh
foreach($module in Get-InstalledModule) {
    Find-Module -Name $module.Name -Repository $module.Repository -MinimumVersion $module.Version | Where Version -ne $module.Version
}
```
### In an elevated host:
8. Use `Update-Module -Whatif` to verify that it detects the module needing updates.
9. Use `Update-Module` to update it.

edyoung:
This changes the behavior of Update-Module and Update-Script so that:

```powershell
Update-Module 
Update-Module -Name *
```
Are equivalent. Both versions will produce a warning for each item found which can't be upgraded due to permissions, instead of silently skipping them. This does mean that some commands will be noisier than they used to be - which is what is requested here but won't necessarily make everyone happy.


alerickson (approved):

alerickson:
This looks good, though I think we should add a test for update-module/script.

edyoung:
This changes the approach: now we mirror the new behavior of Install-Module more closely.

If you have a module or script installed already to AllUsers scope, AND you are using Windows PowerShell AND you are an admin, we will install the new version to AllUsers scope.

Otherwise, we will install it to CurrentUser scope. This will *succeed* (without a warning) even if you are not elevated. 

One downside which was already possible but easier to hit now: you can have 2 versions of a script installed, to AllUsers and CurrentUser scope. Which one you get depends on your $PATH and won't always be the latest one.

PR #428, opened 2019-02-22
Add DSC Unit tests to BVT

mgreenegit:
It appears no one is verifying the PowerShellGet DSC resources.  I would like to request that we add tests, starting with just the basic unit tests, to daily build validation.
THIS IS LIKELY TO BREAK THE BUILD TEMPORARILY.


edyoung:
Build doesn't look broken to me, let's do this

PR #438, opened 2019-03-12
Merge development into master

alerickson:


edyoung:
I don't quite understand why we're seeing all these commits again?

Issue #441, opened 2019-03-14
Register-PSRepository -SourceLocation fails with special chars in path


jborean93:
Steps to reproduce
------------------

```powershell
$path = 'C:\ps repo testing [$!@^&test(;)]'
if (Test-Path -LiteralPath $path) {
    Remove-Item -LiteralPath $path -Force -Recurse
}
New-Item -Path $path -ItemType Directory > $null

Register-PSRepository -Name testing -SourceLocation $path
```

Expected behavior
-----------------
Repo is registered and works as per normal

Actual behavior
---------------
Fails with the following error

```
Register-PSRepository : Cannot find the path 'C:\ps repo testing [$!@^&test(;)]' because it does not exist.
At line:1 char:1
+ Register-PSRepository -Name testing -SourceLocation $path
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (C:\ps repo testing [$!@^&test(;)]:String) [Register-PSRepository], Arg
   umentException
    + FullyQualifiedErrorId : PathNotFound,Register-PSRepository
```

This is because `Resolve-Location` uses `Test-Path -Path` instead of `Test-Path -LiteralPath` and failing with the glob like characters. I can manually change the line at https://github.com/PowerShell/PowerShellGet/blob/development/src/PowerShellGet/private/functions/Resolve-Location.ps1#L32 to use `-LiteralPath` and it's able to continue forward.

After fixing the Resolve-Path issue to use `-LiteralPath` instead of just `-Path`.
```none
PackageManagement\Register-PackageSource : Cannot find the path 'C:\ps repo testing [$!@^&test(;)]' because it does
not exist.
At C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\1.6.0\PSModule.psm1:4631 char:17
+ ...     $null = PackageManagement\Register-PackageSource @PSBoundParamete ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (C:\ps repo testing [$!@^&test(;)]:String) [Register-PackageSource], Ex
   ception
    + FullyQualifiedErrorId : PathNotFound,Add-PackageSource,Microsoft.PowerShell.PackageManagement.Cmdlets.RegisterPa
[...]

PR #442, opened 2019-03-14
Add support for registering repositories with special chars

jborean93:
Fixes https://github.com/PowerShell/PowerShellGet/issues/441.

This allows you to use `Register-PSRepository` to register a file path that contains glob like characters like `[`.


alerickson:
Thanks @jborean93!!

Issue #335, opened 2018-09-25
Refactor Publish-Module into smaller responsibilities


jzabroski:
This code base has gotten long in the tooth and it desperately needs better test coverage and more narrowly defined test cases.

Publish-Module is a good place to start.

One example refactoring: Extract nuspec logic into New-NuspecFile and ConvertTo-NugetPackage

edyoung:
See PR above.

Issue #344, opened 2018-10-08
Publish-Module fails to publish when large number of functions are exported


Texernie:
Publish-Module creates an invalid nuspec file when a large number of functions are exported in a powershell module. The problem is that down in Publish-PSArtifactUtility, it brute forces all of the exported function names into the tags. My particular case resulted in a tag field around 11000 characters, way over the 4000 limit. I am publishing to a self-hosted nuget gallery, but it is based upon the current production branch.

It would be nice if it either truncated the values at 4000 characters with a warning, or allow a total override of the tags.

Steps to reproduce
------------------

```
Create a module with many functions to export. Mine has 116 distinct functions.
$functionNames = (Get-ChildItem '.\MyModule\Scripts' *.ps1).BaseName
Update-ModuleManifest `
    -Path '.\MyModule\MyModule.psd1' `
    -FunctionsToExport $functionNames 

Publish-Module -Path (resolve-path .\MyModule)  | Out-Null

```

Expected behavior
-----------------

Publishing of the module to succeed and a nuget package exist in the repository.

Actual behavior
---------------
```
Publish-PSArtifactUtility : Failed to publish module 'MyModule': 'Response status code does not indicate success: 400 (A nuget package's Tags
property may not be more than 4000 characters long.).
'.
At C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\2.0.1\PSModule.psm1:10859 char:17
+ ...             Publish-PSArtifactUtility @PublishPSArtifactUtility_Param ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : FailedToPublishTheModule,Publish-PSArtifactUtility

```

Environment data
----------------
```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      5.1.17134.228
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17134.228
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

```

```powershell
> Get-Module -ListAvailable PowerShellGet,PackageManagement

[...]

PR #444, opened 2019-03-14
Refactor Nuget operations out of Publish-PSArtifactUtility function

Benny1007:
Refactor nuget operations out of Publish-PSArtifactUtility.ps1 to start #335

### DIFF
--- a/src/PowerShellGet/private/functions/New-NugetPackage.ps1
+++ b/src/PowerShellGet/private/functions/New-NugetPackage.ps1
@@ -1,0 +1,110 @@
+function New-NugetPackage {
+    [CmdletBinding()]
+    Param(
+        [Parameter(Mandatory = $true)]
+        [string]$NuspecPath,
+
+        [Parameter(Mandatory = $true)]
+        [string]$NugetPackageRoot,
+
+        [Parameter()]
+        [string]$OutputPath = $NugetPackageRoot,
+
+        [Parameter(ParameterSetName = "UseNuget")]
+        [string]$NugetExePath,
+
+        [Parameter(ParameterSetName = "UseDotnetCli")]
+        [switch]$UseDotnetCli
+
+    )
+
+    if (-Not(Test-Path -Path $NuspecPath -PathType Leaf)) {
+        throw "A nuspec file does not exist at $NuspecPath, provide valid path to a .nuspec"
+    }
+
+    if (-Not(Test-Path -Path $NugetPackageRoot)) {
+        throw "NugetPackageRoot $NugetPackageRoot does not exist"
+    }
+
+    if ($PSCmdlet.ParameterSetName -eq "UseNuget") {
+        if (-Not(Test-Path -Path $NuGetExePath)) {
+            throw "Nuget.exe does not exist at $NugetExePath, provide a valid path to nuget.exe"
+        }
+
+        $ArgumentList = @("pack")
+        $ArgumentList += "`"$NuspecPath`""
+        $ArgumentList += "-outputdirectory `"$OutputPath`""
+
+        $processStartInfo = New-Object System.Diagnostics.ProcessStartInfo
+        $processStartInfo.FileName = $NugetExePath
+        $processStartInfo.RedirectStandardError = $true
+        $processStartInfo.RedirectStandardOutput = $true
+        $processStartInfo.UseShellExecute = $false
+        $processStartInfo.Arguments = $ArgumentList
+
+        $process = New-Object System.Diagnostics.Process
+        $process.StartInfo = $processStartInfo
+        $process.Start() | Out-Null
+        $process.WaitForExit()
+
+        if (-Not ($process.ExitCode -eq 0 )) {
+            $stdErr = $process.StandardError.ReadToEnd()
+            throw "nuget.exe failed to pack $stdErr"
+        }
+    }
+
+    if ($PSCmdlet.ParameterSetName -eq "UseDotnetCli") {
+        #perform dotnet pack using a temporary project file.
+        $dotnetCliPath = (Get-Command -Name "dotnet").Source
+        $tempPath = Join-Path -Path ([System.IO.Path]::GetTempPath()) -ChildPath ([System.Guid]::NewGuid()).Guid
+        New-Item -ItemType Directory -Path $tempPath -Force | Out-Null
+
+        $CsprojContent = @"
+            <Project Sdk="Microsoft.NET.Sdk">
+            <PropertyGroup>
+                <AssemblyName>NotUsed</AssemblyName>
+                <Description>Temp project used for creating nupkg file.</Description>
+                <TargetFramework>netcoreapp2.0</TargetFramework>
+                <IsPackable>true</IsPackable>
+            </PropertyGroup>
+            </Project>
+"@
+        $projectFile = New-Item -ItemType File -Path $tempPath -Name "Temp.csproj"
+        Set-Content -Value $CsprojContent -Path $projectFile
+
+        #execution
+
+        $ArgumentList = @("pack")
+        $ArgumentList += "`"$projectFile`""
+        $ArgumentList += "/p:NuspecFile=`"$NuspecPath`""
+        $ArgumentList += "--output `"$OutputPath`""
+
+        $processStartInfo = New-Object System.Diagnostics.ProcessStartInfo
+        $processStartInfo.FileName = $dotnetCliPath
+        $processStartInfo.RedirectStandardError = $true
+        $processStartInfo.RedirectStandardOutput = $true
+        $processStartInfo.UseShellExecute = $false
+        $processStartInfo.Arguments = $ArgumentList
+
+        $process = New-Object System.Diagnostics.Process
+        $process.StartInfo = $processStartInfo
+        $process.Start() | Out-Null
+        $process.WaitForExit()
+
+        if (-Not ($process.ExitCode -eq 0 )) {
+            if (Test-Path -Path $tempPath) {
+                Remove-Item -Path $tempPath -Force -Recurse
+            }
+            $stdOut = $process.StandardOut.ReadToEnd()
+            throw "dotnet cli failed to pack $stdOut"
+        }
+
+        if (Test-Path -Path $tempPath) {
+            Remove-Item -Path $tempPath -Force -Recurse
+        }
+    }
+
+    $stdOut = $process.StandardOutput.ReadLine()
+
+    Write-Verbose -Message $stdOut
+}
--- a/src/PowerShellGet/private/functions/New-NugetPackage.ps1
+++ b/src/PowerShellGet/private/functions/New-NugetPackage.ps1
@@ -91,17 +91,15 @@ function New-NugetPackage {
         $process.Start() | Out-Null
         $process.WaitForExit()
 
+        if (Test-Path -Path $tempPath) {
+            Remove-Item -Path $tempPath -Force -Recurse
+        }
+
         if (-Not ($process.ExitCode -eq 0 )) {
-            if (Test-Path -Path $tempPath) {
-                Remove-Item -Path $tempPath -Force -Recurse
-            }
             $stdOut = $process.StandardOut.ReadToEnd()
             throw "dotnet cli failed to pack $stdOut"
         }
 
-        if (Test-Path -Path $tempPath) {
-            Remove-Item -Path $tempPath -Force -Recurse
-        }
     }
 
     $stdOut = $process.StandardOutput.ReadLine()
--- a/src/PowerShellGet/private/functions/New-NugetPackage.ps1
+++ b/src/PowerShellGet/private/functions/New-NugetPackage.ps1
@@ -17,6 +17,7 @@ function New-NugetPackage {
         [switch]$UseDotnetCli
 
     )
+    Set-StrictMode -Off
 
     if (-Not(Test-Path -Path $NuspecPath -PathType Leaf)) {
         throw "A nuspec file does not exist at $NuspecPath, provide valid path to a .nuspec"
--- a/src/PowerShellGet/private/functions/New-NugetPackage.ps1
+++ b/src/PowerShellGet/private/functions/New-NugetPackage.ps1
@@ -104,10 +104,9 @@ function New-NugetPackage {
 
     }
 
-    [xml]$nuspecXml = Get-Content -Path $NuspecPath
-    $version = $nuspecXml.package.metadata.version
-    $id = $nuspecXml.package.metadata.id
-    $nupkgFullFile = Join-Path $OutputPath -ChildPath "$id.$version.nupkg"
+    $stdOut = $process.StandardOutput.ReadToEnd()
+    $stdOut -match "Successfully created package '(.*.nupkg)'" | Out-Null
+    $nupkgFullFile = $matches[1]
 
     $stdOut = $process.StandardOutput.ReadToEnd()
 
--- a/src/PowerShellGet/private/functions/New-NugetPackage.ps1
+++ b/src/PowerShellGet/private/functions/New-NugetPackage.ps1
@@ -10,7 +10,7 @@ function New-NugetPackage {
         [Parameter()]
         [string]$OutputPath = $NugetPackageRoot,
 
-        [Parameter(ParameterSetName = "UseNuget")]
+        [Parameter(Mandatory = $true, ParameterSetName = "UseNuget")]
         [string]$NugetExePath,
 
         [Parameter(ParameterSetName = "UseDotnetCli")]
@@ -25,6 +25,7 @@ function New-NugetPackage {
     if (-Not(Test-Path -Path $NugetPackageRoot)) {
         throw "NugetPackageRoot $NugetPackageRoot does not exist"
     }
+
 
     if ($PSCmdlet.ParameterSetName -eq "UseNuget") {
         if (-Not(Test-Path -Path $NuGetExePath)) {
@@ -96,13 +97,19 @@ function New-NugetPackage {
         }
 
         if (-Not ($process.ExitCode -eq 0 )) {
-            $stdOut = $process.StandardOut.ReadToEnd()
+            $stdOut = $process.StandardOutput.ReadToEnd()
             throw "dotnet cli failed to pack $stdOut"
         }
 
     }
 
-    $stdOut = $process.StandardOutput.ReadLine()
+    [xml]$nuspecXml = Get-Content -Path $NuspecPath
+    $version = $nuspecXml.package.metadata.version
+    $id = $nuspecXml.package.metadata.id
+    $nupkgFullFile = Join-Path $OutputPath -ChildPath "$id.$version.nupkg"
+
+    $stdOut = $process.StandardOutput.ReadToEnd()
 
     Write-Verbose -Message $stdOut
+    Write-Output $nupkgFullFile
 }
--- a/src/PowerShellGet/private/functions/New-NuspecFile.ps1
+++ b/src/PowerShellGet/private/functions/New-NuspecFile.ps1
@@ -47,6 +47,7 @@ function New-NuspecFile {
         [PSObject[]]$Files
 
     )
+    Set-StrictMode -Off
 
     $nameSpaceUri = "http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd"
     [xml]$xml = New-Object System.Xml.XmlDocument
--- a/src/PowerShellGet/private/functions/New-NuspecFile.ps1
+++ b/src/PowerShellGet/private/functions/New-NuspecFile.ps1
@@ -62,8 +62,7 @@ function New-NuspecFile {
     $Tags = $Tags -Join " "
 
     if ($Tags.Length -gt 4000) {
-        $Tags = $Tags.Substring(0, $Tags.LastIndexOf(" "))
-        Write-Warning -Message "Nuspec 'Tag' list exceeded max 4000 characters and was truncated."
+        Write-Warning -Message "Tag list exceeded 4000 characters and may not be accepted by some Nuget feeds."
     }
 
     $metaDataElementsHash = [ordered]@{
@@ -121,5 +120,8 @@ function New-NuspecFile {
 
     $xml.AppendChild($packageElement) | Out-Null
 
-    $xml.save("$OutputPath\$Id.nuspec")
+    $nuspecFullName = Join-Path -Path $OutputPath -ChildPath "$Id.nuspec"
+    $xml.save($nuspecFullName)
+
+    Write-Output $nuspecFullName
 }
--- a/src/PowerShellGet/private/functions/New-NuspecFile.ps1
+++ b/src/PowerShellGet/private/functions/New-NuspecFile.ps1
@@ -77,9 +77,9 @@ function New-NuspecFile {
         tags                     = $Tags
     }
 
-    if ($LicenseUrl) { $metaDataElement.Add("licenseUrl", $LicenseUrl) }
-    if ($ProjectUrl) { $metaDataElement.Add("projectUrl", $ProjectUrl) }
-    if ($IconUrl) { $metaDataElement.Add("iconUrl", $IconUrl) }
+    if ($LicenseUrl) { $metaDataElementsHash.Add("licenseUrl", $LicenseUrl) }
+    if ($ProjectUrl) { $metaDataElementsHash.Add("projectUrl", $ProjectUrl) }
+    if ($IconUrl) { $metaDataElementsHash.Add("iconUrl", $IconUrl) }
 
     foreach ($key in $metaDataElementsHash.Keys) {
         $element = $xml.CreateElement($key, $nameSpaceUri)
@@ -88,6 +88,7 @@ function New-NuspecFile {
 
         $metaDataElement.AppendChild($element) | Out-Null
     }
+
 
     if ($Dependencies) {
         $dependenciesElement = $xml.CreateElement("dependencies", $nameSpaceUri)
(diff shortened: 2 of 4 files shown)


src/PowerShellGet/private/functions/New-NugetPackage.ps1
```diff
@@ -0,0 +1,110 @@
...
+
+        if (-Not ($process.ExitCode -eq 0 )) {
+            $stdErr = $process.StandardError.ReadToEnd()
+            throw "nuget.exe failed to pack $stdErr"
+        }
+    }
+
+    if ($PSCmdlet.ParameterSetName -eq "UseDotnetCli") {
+        #perform dotnet pack using a temporary project file.
+        $dotnetCliPath = (Get-Command -Name "dotnet").Source
```
edyoung, line 60: for symmetry with the nuget.exe case we shouldn't we  check if dotnet was  found?

src/PowerShellGet/private/functions/New-NugetPackage.ps1
```diff
@@ -0,0 +1,110 @@
...
+        $processStartInfo.RedirectStandardError = $true
+        $processStartInfo.RedirectStandardOutput = $true
+        $processStartInfo.UseShellExecute = $false
+        $processStartInfo.Arguments = $ArgumentList
+
+        $process = New-Object System.Diagnostics.Process
+        $process.StartInfo = $processStartInfo
+        $process.Start() | Out-Null
+        $process.WaitForExit()
+
```
edyoung, line 95: why not remove the temp path here instead of having the code twice below

src/PowerShellGet/private/functions/New-NuspecFile.ps1
```diff
@@ -0,0 +1,125 @@
...
+    $packageElement = $xml.CreateElement("package", $nameSpaceUri)
+    $metaDataElement = $xml.CreateElement("metadata", $nameSpaceUri)
+
+    #truncate tags if they exceed nuspec specifications for size.
+    $Tags = $Tags -Join " "
+
+    if ($Tags.Length -gt 4000) {
+        $Tags = $Tags.Substring(0, $Tags.LastIndexOf(" "))
+        Write-Warning -Message "Nuspec 'Tag' list exceeded max 4000 characters and was truncated."
+    }
```
edyoung, line 67: PowerShellGallery does not enforce the 4000 character limit and we have modules which exceed it. Probably better to Write-Warning "Tag list exceeded 4000 characters and may not be accepted by some Nuget feeds" without truncating it

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
```diff
@@ -372,267 +329,116 @@ function Publish-PSArtifactUtility
...
-"@
-    $NupkgPath = Microsoft.PowerShell.Management\Join-Path -Path $NugetPackageRoot -ChildPath "$Name.$Version.nupkg"
-
-    $csprojBasePath = $null
-    if($script:DotnetCommandPath) {
-        $csprojBasePath = Microsoft.PowerShell.Management\Join-Path -Path $script:TempPath -ChildPath ([System.Guid]::NewGuid())
-        $null = Microsoft.PowerShell.Management\New-Item -Path $csprojBasePath -ItemType Directory -Force -WhatIf:$false -Confirm:$false
-        $NuspecPath = Microsoft.PowerShell.Management\Join-Path -Path $csprojBasePath -ChildPath "$Name.nuspec"
-        $CsprojFilePath = Microsoft.PowerShell.Management\Join-Path -Path $csprojBasePath -ChildPath "$Name.csproj"
+        $dependencyObject = New-Object -TypeName PSCustomObject -Property $props
```
edyoung, line 350: You could also do $dependencyObject = [PSCustomObject]$props, btw

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
```diff
@@ -372,267 +329,116 @@ function Publish-PSArtifactUtility
...
-        }
-        elseif($script:NuGetExePath) {
-            $output = & $script:NuGetExePath pack $NuspecPath -OutputDirectory $NugetPackageRoot
-        }
-
-        if(-not (Test-Path -Path $NupkgPath -PathType Leaf)) {
-            $SemanticVersionString = Get-NormalizedVersionString -Version $Version
-            $NupkgPath = Join-PathUtility -Path $NugetPackageRoot -ChildPath "$Name.$($SemanticVersionString).nupkg" -PathType File
-        }
+    # fix#335 refactor to use new New-NuspecFile function.
```
edyoung: I don't think we need the comment, it won't mean anything to a future reader of the code

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
```diff
@@ -372,267 +329,116 @@ function Publish-PSArtifactUtility
...
-                $errorId = "FailedToCreateCompressedModule"
-            }
-            else
-            {
-                $message = $LocalizedData.FailedToCreateCompressedScript -f ($output)
-                $errorId = "FailedToCreateCompressedScript"
-            }
+    try {
+        New-NuspecFile $params
+        $NuspecFullName = Join-Path -Path $NugetPackageRoot -ChildPath "$Name.nuspec"
```
edyoung: You could consider having New-NuspecFile return the path of the file it created? up to you

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
```diff
@@ -372,267 +329,116 @@ function Publish-PSArtifactUtility
...
-        $outputMsg = $null
-        $StartProcess_params = @{
-            RedirectStandardError = $tempErrorFile
-            RedirectStandardOutput = $tempOutputFile
-            NoNewWindow = $true
-            Wait = $true
-            PassThru = $true
+        elseif ($NuGetExePath) {
+            New-NugetPackage -NuspecPath $NuspecFullName -NugetPackageRoot $NugetPackageRoot -NugetExePath $NuGetExePath -Verbose:$VerbosePreference
         }
```
edyoung, line 384: does there need to be an else to cover producing an error when neither nuget or dotnet were found?

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
```diff
@@ -372,267 +329,116 @@ function Publish-PSArtifactUtility
...
-        }
-        elseif($script:NuGetExePath) {
-            $StartProcess_params['FilePath'] = $script:NuGetExePath
-
-            $ArgumentList = @('push')
-            $ArgumentList += "`"$NupkgPath`""
-            $ArgumentList += @('-source', "`"$($Destination.TrimEnd('\'))`"")
-            $ArgumentList += @('-apikey', "`"$NugetApiKey`"")
-            $ArgumentList += '-NonInteractive'
+        $NupkgFullName = Join-Path -Path $NugetPackageRoot -ChildPath "$Name.$Version.nupkg"
```
edyoung: New-NugetPackage could return name of the package it created

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
```diff
@@ -372,267 +329,116 @@ function Publish-PSArtifactUtility
...
-
-        if(Test-Path -Path $tempErrorFile -PathType Leaf) {
-            $errorMsg = Microsoft.PowerShell.Management\Get-Content -Path $tempErrorFile -Raw
+        Write-Error -Message $message -ErrorId $errorId -Category InvalidOperation -ErrorAction Stop
+    }
 
-            if($errorMsg) {
-                Write-Verbose -Message $errorMsg
-            }
+    # Fix#335 refactor to use new Publish-NugetPackage function.
```
edyoung: I'm not seeing Publish-NugetPackage in this PR?

edyoung (changes requested):
Thanks a lot - I think this will definitely be cleaner than current state. Not all of the comments (especially the ones which say 'consider') are mandatory - if you don't think they make sense just resolve them

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
Benny1007: Oh wow yeah, have re-added.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
Benny1007: We could do for symmetry I guess, however NugetExePath is fed in as an input, the dotnet cli command is found using Get-Command and indexing into .Source on line 58 would throw an error which would be picked up by the try/catch in line 382 of the calling Publish-PSArtifactUtility function. Personally I prefer this approach and would rather have the nuget.exe case call get-command nuget.exe or something similar but I am not sure of all the ramifications and where else exactly nuget.exe is searched for and found using PowershellGet. Happy to change though unless you can shed any more light on it.

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
Benny1007: line 69 should populate these script level variables though with Install-NuGetClientBinaries. Not a fan of that tbh though. If those variables aren't set would you prefer to exit gracefully with an error message? Would Install-NuGetClientBinaries not do that? not looked at that code.

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
Benny1007: I think i'll leave this if that's ok, I'd rather explicitly check the file exists and have that blow up catching the exception than assign the result to $NuspecFullName then test-path in an if.

Benny1007: actually scratch that i've made that change.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
Benny1007: ah yep makes sense. done.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
Benny1007: Done.

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
Benny1007: not done

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
Benny1007: done.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
edyoung: OK

Benny1007:
Have addressed most feedback. I can't seem to get these tests to even complete locally or on the build servers though. Do you have any tips? - The test suite is pretty scary.

alerickson:
@Benny1007 check out https://github.com/PowerShell/PowerShellGet#running-tests for running tests locally.  You can also run Invoke-Pester with a particular test file instead of running the entire suite. 
Also be sure to first run "import-module src/PowerShellGet -force" to ensure you have your latest changes before running tests.

Benny1007:
> @Benny1007 check out https://github.com/PowerShell/PowerShellGet#running-tests for running tests locally. You can also run Invoke-Pester with a particular test file instead of running the entire suite.
> Also be sure to first run "import-module src/PowerShellGet -force" to ensure you have your latest changes before running tests.

Thanks, so I have run the entire suite locally and everything passes. 
I have also run the entire suite on my macbook with the latest powershell, this seems to have other issues in some other tests related to permissions to temp drive. 

I have no idea why they are not working on the build server, but it's like they never even finish.

edyoung:
The tests fail because you're calling a function with a mandatory param but not passing a value. Powershell then stops everything and prompts for the user to supply the argument, which isn't so useful on a CI server, then times out waiting. See this snippet from Travis CI log:

```
cmdlet New-NuspecFile at command pipeline position 1

Supply values for the following parameters:

Id: 


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

Benny1007:
Cool thanks. It's concerning none of this was picked up running the test suite locally. So within the calling function, we can't rely on $PSModuleInfo.Name or $PSScriptInfo.Name having a value or do we just need to fix up the test data where it creates these modules/scripts?

Benny1007:
Something that has led to false positives during testing is that if I unload PowerShellGet module with:-
`Remove-Module -Name PowerShellGet -Force`
then load it from the code base with 
`Import-Module -Name .\src\PowerShellGet\PowerShellGet.psd1 -Force`
Then break on say line 119 of PSGet.PublishModule.Tests
I then seem to have 2 PowerShellGet modules loaded. I think this is caused by line 1160 in PSGetTestUtils.psm1
This caused the newer PowerShellGet from C:\Program Files\WindowsPowerShell\Modules\PowerShellGet version 2.1.2 to be loaded and used hence forth masking my issues. 
You come across that before?

alerickson:
@Benny1007 no I haven't.  Are you following option 1 or option 2 when running tests (https://github.com/PowerShell/PowerShellGet#running-tests)?  I've had issues with the env path unintentionally changing and importing the wrong module, so I prefer to follow the latter.

edyoung:
I always set $env:psmodulepath to pick up the version under development rather than the installed one, that avoids issues like that

Benny1007:
As mentioned in the PR feedback truncating of the Tags has been removed and replace with a warning which won't exactly help #344 but will provided a clearer reason why the publish has failed?

awickham10:
> As mentioned in the PR feedback truncating of the Tags has been removed and replace with a warning which won't exactly help #344 but will provided a clearer reason why the publish has failed?

It seems like the problem is Publish-Module includes every function as a tag. What if a switch was added to exclude functions as tags? Something like... `Publish-Module -Name 'MyModule' -ExcludeFunctionsAsTags` that way the odds of hitting the 4000 character limit is drastically reduced?

edyoung (approved):

edyoung:
@awickham10 I'd be fine with that (or a -EnforceMaximumTagLength option maybe), but I'd rather keep that a separate change so this is a pure refactor

edyoung:
Thanks @Benny1007 !

alerickson:
The Visual Studio 2017, PowerShell Core environment on AppVeyor says it's passing but didn't actually run.  See: https://ci.appveyor.com/project/PowerShell/powershellget/builds/23226384/job/hu9xotrcnsvkyb8a

Benny1007:
That's not very good. The build server reports success on a failure?
Looks like the build server didn't have PowerShell core (pwsh.exe) installed?

src/PowerShellGet/private/functions/New-NugetPackage.ps1
jzabroski: @Benny1007 @edyoung I would like a Verbose Preference to enable logging which Get-Command is retrieved.  What do you think?

src/PowerShellGet/private/functions/New-NugetPackage.ps1
```diff
@@ -0,0 +1,110 @@
...
+        if (-Not ($process.ExitCode -eq 0 )) {
+            $stdErr = $process.StandardError.ReadToEnd()
+            throw "nuget.exe failed to pack $stdErr"
+        }
+    }
+
+    if ($PSCmdlet.ParameterSetName -eq "UseDotnetCli") {
+        #perform dotnet pack using a temporary project file.
```
jzabroski, line 59: @edyoung  @Benny1007  Extract this into it's own function.  Replace the comment with a function name.  Use a PowerShell approved verb.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
jzabroski, line 60: Create Verbose Preference to log which Get-Command is retrieved.  This would be useful in submitting bug reports.  It could even verbose the version of the exe found.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
```diff
@@ -0,0 +1,110 @@
...
+        $process.StartInfo = $processStartInfo
+        $process.Start() | Out-Null
+        $process.WaitForExit()
+
+        if (-Not ($process.ExitCode -eq 0 )) {
+            if (Test-Path -Path $tempPath) {
+                Remove-Item -Path $tempPath -Force -Recurse
+            }
+            $stdOut = $process.StandardOut.ReadToEnd()
+            throw "dotnet cli failed to pack $stdOut"
```
jzabroski, line 102: This still does not allow inspecting the temporary created project file.  This makes it a bit magical when trying to understand what got uploaded to a repository as part of the publish process.

@Benny1007 @edyoung

jzabroski:
@edyoung @Benny1007  I just saw this commit - for some reason it showed up in my GitHub Read folder instead of notifications.  I only reviewed one file so far, though.

jzabroski (commented):
@alerickson  Are you saying @Benny1007  broke the build, or someone in charge of the Appveyor commands wrote a bad pipeline?  Or just "this is broke, I hope someone fixes it"?

alerickson:
@jzabroski just meant that the PR shouldn't have been merged because the tests for core didn't run.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
```diff
@@ -0,0 +1,125 @@
...
+    #create top-level elements
+    $packageElement = $xml.CreateElement("package", $nameSpaceUri)
+    $metaDataElement = $xml.CreateElement("metadata", $nameSpaceUri)
+
+    #truncate tags if they exceed nuspec specifications for size.
+    $Tags = $Tags -Join " "
+
+    if ($Tags.Length -gt 4000) {
+        $Tags = $Tags.Substring(0, $Tags.LastIndexOf(" "))
+        Write-Warning -Message "Nuspec 'Tag' list exceeded max 4000 characters and was truncated."
```
jzabroski: This appears to address https://github.com/PowerShell/PowerShellGet/issues/344

Additionally, now that this is a first-class function exposable to end-users, the additional request to make 4000 limit configurable for users who want to upload packages to their own nuget server is possible

src/PowerShellGet/private/functions/New-NuspecFile.ps1
```diff
@@ -0,0 +1,125 @@
...
+    [xml]$xml = New-Object System.Xml.XmlDocument
+
+    $xmlDeclaration = $xml.CreateXmlDeclaration("1.0", "utf-8", $null)
+    $xml.AppendChild($xmlDeclaration) | Out-Null
+
+    #create top-level elements
+    $packageElement = $xml.CreateElement("package", $nameSpaceUri)
+    $metaDataElement = $xml.CreateElement("metadata", $nameSpaceUri)
+
+    #truncate tags if they exceed nuspec specifications for size.
```
jzabroski, line 62: If you're destructively messing with parameters, create a Process block attached to the params block.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
```diff
@@ -0,0 +1,125 @@
...
+    $Tags = $Tags -Join " "
+
+    if ($Tags.Length -gt 4000) {
+        $Tags = $Tags.Substring(0, $Tags.LastIndexOf(" "))
```
jzabroski: This is not guaranteed to truncate at the lesser of 4,000 characters or "full list of tags separated by spaces up to 4,000 characters"

src/PowerShellGet/private/functions/New-NuspecFile.ps1
jzabroski: @Benny1007  The message should be updated to say you're not truncating it.

src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
jzabroski: I think the style @Benny1007  wrote here is more readable.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
Benny1007: what are the benefits of abstracting this further. Don't want to get into similar situation of Register-PSRepository and Resolve-Location, Get-LocationString, Get-EscapedString etc... This function has a single responsibility - to create a nuget package, having all the code at hand in one file to perform that was the reason for abstracting it in the first place. Personally not a fan of splitting out to multiple functions for the sake of it.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
Benny1007: Overwrite the outputpath variable and then you would know where the artifact resides.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
Benny1007: can you explain this further?

src/PowerShellGet/private/functions/New-NuspecFile.ps1
Benny1007: These nuget functions are still private functions and not exposed in the module to the public.

jzabroski:
How am I supposed to overwrite the output path on a private function?

jzabroski:
What's the benefit of having nuget.exe instead of nugetupdate.exe,
nugetpublish.exe, nugethelp.exe? Reductio ad absurdum.

If you think my suggestion makes the code less readable, that's OK. But I
actually think it would improve readability since all you really care about
is "do something, aggregate stdout, srderr and exit code, and return." I
think the fact it's already not handled consistently across the code base
is reason enough to add a helper. Call it Helper/Invoke-Process so that
it's clearer its a cross cutting concept. At the end of the day you should
not care about the ceremony of Process start, redirection and exit codes.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
jzabroski: In PowerShell v1, the only way to validate parameters was in a Process block and to throw an exception.  In PowerShell v2, you can use ValidateScript.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
jzabroski: @Benny1007  Here is a C# way to do what I'm describing. The author is kind enough to post sample data that you can use to test against.  I also recommend making this a private helper function for readability: https://codereview.stackexchange.com/a/163788/193241

src/PowerShellGet/private/functions/New-NuspecFile.ps1
jzabroski: Then how am I supposed to follow your advice to overwrite the output path?   See your comment here https://github.com/PowerShell/PowerShellGet/pull/444#discussion_r269376413 - maybe I'm missing something obvious and I apologize for being stubborn if that's the case.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
jzabroski: It's not for the sake of it.  It's for building reusable abstractions.  Why is it out fault that the Process API is not designed for consuming a quick task's exit code and standard output and error streams?  We just wrap that and clearly express we care about those 3 things.  See my comment here which when I replied by email 2 days ago somehow got stuck at the bottom of the review: https://github.com/PowerShell/PowerShellGet/pull/444#issuecomment-477119239

jzabroski:
> @awickham10 I'd be fine with that (or a -EnforceMaximumTagLength option maybe), but I'd rather keep that a separate change so this is a pure refactor

This is a private function right now so it doesn't add a lot of value for something that will never have code coverage in production usage.  The more important question is whether some of these helper functions should be public instead of private.  My gut feeling is that they should be split out into a separate helper module, so that if people want to debug PowerShellGet internals, they can.  This is not a traditional way of thinking about using modules, but it just "Feels right" to me.  Also, by making it a separate module instead of public interface to a major mainstream package, it allows for easier user story to sunset the module once PowerShellGet becomes more stable over time and the need for these utility functions goes away from the public eye.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
Benny1007: The snippet is outdated, it was recommended in the PR feedback to NOT truncate.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
Benny1007: There would have to be another separate PR that exposed the outputpath via the public members. This small refactor was just the start of the re-write of Publish-Module I assume the refactoring for this module will be across Publish-PSArtifactUtility (potentially going away) and Publish-Module.

src/PowerShellGet/private/functions/New-NugetPackage.ps1
Benny1007: sorry, was missing a question mark off the last comment! would need to pass this through from the public function.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
Benny1007: I think this comment was in reference to the truncation of tags?, which is now NOT happening but I seem to have left the comment in the code.

PR #452, opened 2019-03-22
Add SkipAutomaticTags Parameter to Publish-Module

awickham10:
When using a generic NuGet Server there is a hard coded 4000 character limit. Publish-Module gets a list of exported functions from the module manifest. If there are a lot of functions exported, publishing to a generic NuGet Server fails with an error stating: A nuget package's Tags property may not be more than 4000 characters long. This PR is to resolve issue #344.

The -EnforceMaximumTagLength switch removes commands and resources from being included as tags when present. This switch was added to Publish-Module and Publish-PSArtificateUtility.

I'm open to discussion on how the switch functions. I don't really have much of a preference - I just need the module to publish successfully.

I also wasn't sure where the help documentation was located, so that still needs to be added.

### DIFF
--- a/src/PowerShellGet/private/functions/New-NuspecFile.ps1
+++ b/src/PowerShellGet/private/functions/New-NuspecFile.ps1
@@ -59,10 +59,9 @@ function New-NuspecFile {
     $packageElement = $xml.CreateElement("package", $nameSpaceUri)
     $metaDataElement = $xml.CreateElement("metadata", $nameSpaceUri)
 
-    #truncate tags if they exceed nuspec specifications for size.
-    $Tags = $Tags -Join " "
-
-    if ($Tags.Length -gt 4000) {
+    # warn we're over 4000 characters for standard nuget servers
+    $tagsString = $Tags -Join " "
+    if ($tagsString.Length -gt 4000) {
         Write-Warning -Message "Tag list exceeded 4000 characters and may not be accepted by some Nuget feeds."
     }
 
--- a/src/PowerShellGet/private/functions/New-NuspecFile.ps1
+++ b/src/PowerShellGet/private/functions/New-NuspecFile.ps1
@@ -74,7 +74,7 @@ function New-NuspecFile {
         releaseNotes             = $ReleaseNotes
         requireLicenseAcceptance = $RequireLicenseAcceptance.ToString().ToLower()
         copyright                = $Copyright
-        tags                     = $Tags
+        tags                     = $tagsString
     }
 
     if ($LicenseUrl) { $metaDataElementsHash.Add("licenseUrl", $LicenseUrl) }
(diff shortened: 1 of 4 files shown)


edyoung:
Thanks @awickham10 ! Looking at this a bit, I think that the parameter name may be misleading - the effect is not to enforce the parameter length but to skip automatically adding tags to the module. So maybe we could rename the parameter. How about -SkipAutomaticTags? (Sorry, I think the current name might have been my suggestion.)

src/PowerShellGet/private/functions/New-NuspecFile.ps1
```diff
@@ -59,10 +59,9 @@ function New-NuspecFile {
     $packageElement = $xml.CreateElement("package", $nameSpaceUri)
     $metaDataElement = $xml.CreateElement("metadata", $nameSpaceUri)
 
-    #truncate tags if they exceed nuspec specifications for size.
-    $Tags = $Tags -Join " "
-
-    if ($Tags.Length -gt 4000) {
+    # warn we're over 4000 characters for standard nuget servers
+    $tagsString = $Tags -Join " "
```
Benny1007, line 63: Is there a need to assign to another variable here? Tags is now left as an array and when we write it out to the nuspec file it may just appear as System.Object[] or something.

awickham10: During my testing it seems that since Tags was a parameter defined as a string array it was always kept as an array during the execution. Tags.Length always ended up with a length of 1 (1 element in the array).

awickham10:
> Thanks @awickham10 ! Looking at this a bit, I think that the parameter name may be misleading - the effect is not to enforce the parameter length but to skip automatically adding tags to the module. So maybe we could rename the parameter. How about -SkipAutomaticTags? (Sorry, I think the current name might have been my suggestion.)

SkipAutomaticTags feels a lot better. 

I was thinking more on this and I almost wonder if this shouldn't be a setting on the repository. I'm thinking we want people to have a consistent experience on powershellgallery.com and having this as a parameter on Publish-Module people may be more apt to changing it. If it's a default on the repository people may leave it alone unless they absolutely need it.

src/PowerShellGet/private/functions/New-NuspecFile.ps1
awickham10: Good point though. I'm going to change this PR to WIP and make sure to validate that further.

jzabroski: @awickham10  Please see my comments to @Benny1007  on his old PR.  https://codereview.stackexchange.com/a/163788/193241

edyoung:
I actually prefer this over a flag on the repo. If we did that we would still need some way to manage the set of tags in case they are longer than the limit. Let me know if you are ok with merging this.

jzabroski:
The whole tags thing is completely nonsense anyway you slice it. It's just public functions. The ONLY TIME this way of searching helped me was finding an obscure Active Directory package called HAWK written by a Microsoft Security engineer.

edyoung:
@awickham10  any chance you can change the param name, then we can merge?

awickham10:
> @awickham10 any chance you can change the param name, then we can merge?

Renamed EnforceMaximumTagLength to SkipAutomaticTags.

edyoung (approved):
Thanks!

potatoqualitee:
Sweet, added -SkipAutomaticTags  to my publish routine. Thanks @awickham10 !

PR #462, opened 2019-04-11
WIP: localdev publishing

Benny1007:
For discussion...

Amendments to Build.psm1
- Add LocalDevInstall switch to Install-PublishedModule
- Logic to install the module from ./dist to powershell modules path under version.9999 e.g. 2.1.2.9999
- Add function Install-DevelopmentModule that will call this function with the -LocalDevInstall switch

Amendments to .vscode file
- Add task to run Install-Dependencies
- Add task to run Install-DevelopmentModule
- Add task to run Uninstall-DevelopmentModule
- Add task to run Invoke-PowerShellGetTest

Amendments to readme.md
- Update help to explain above. 

Notes:-
Only the module version in the psd1 file in ./dist is updated with the .9999 - this will never be checked in.
Build server will never run with the -LocalDevInstall switch.
The zip artifact published to ./dist is never modified.


edyoung:
Thanks @Benny1007 . I haven't tried this but looks reasonable. Could you look at the Travis CI test failures to see if they may be related to the change? 

BTW I typically change $env:psmodulepath to include $pwd\src rather than installing the module while working on it.

Issue #143, opened 2017-06-19
Missing NuGetApiKey should not throw exceptions on Publish-Module


Jaykul:
The NuGetApiKey is not **marked** mandatory on Publish-Module.

In point of fact, there are NuGet repositories, like the Visual Studio Online package repositories, which require credentials _rather than_ API keys, so this error (with id "NuGetApiKeyIsRequiredForNuGetBasedGalleryService") is technically incorrect anyway.

You should just try the publish and provide proper errors _when and if_ the authentication fails.

bmanikm:
I agree with you.

jzabroski:
Now that PowerShell/PowerShellGetv2#335 is refactored thanks to @Benny1007 , this should be re-considered

heidarj:
Hey, I would like to try my hand at making an open source contribution, shouldn't this be as simple removing the _"[ValidateNotNullOrEmpty()]"_ from the **NuGetApiKey** parameter and then replacing this:
```powershell

if (-not $NuGetApiKey.Trim()) {
	if (Microsoft.PowerShell.Management\Test-Path -Path $DestinationLocation) {
	$NuGetApiKey = "$(Get-Random)"
	}
	else {
		$message = $LocalizedData.NuGetApiKeyIsRequiredForNuGetBasedGalleryService -f ($Repository, $DestinationLocation)
		ThrowError -ExceptionName "System.ArgumentException" `
			-ExceptionMessage $message `
			-ErrorId "NuGetApiKeyIsRequiredForNuGetBasedGalleryService" `
			-CallerPSCmdlet $PSCmdlet `
			-ErrorCategory InvalidArgument
	}
}
```
with this:
```powershell
if (-not $NuGetApiKey.Trim()) {
	$NuGetApiKey = "$(Get-Random)"
}
```

As I stated I am not familiar with the open source contribution workflow, but I have forked the repo and have tested the change on my machine and it seems to work, I am currently trying to run the tests but am not familiar with pester and testing in general.

jzabroski:
No,  You should require credentials _or_ a NuGetApiKey, or possibly both.  Can you explain why you would pass a random api key?  It's interesting that you report it works - what happens if you don't pass any key?

In our environment, we use Idea ProGet.  They explain their rational for not using NuGetApiKey here: https://inedo.com/support/documentation/proget/feeds/nuget#nuget

> # NuGet API Key
[...]

PR #463, opened 2019-04-12
NuGetApiKey is required for nuget based gallery service

heidarj:
Hey, I would like to submit my attempt at fixing the problem described in issue #143

I removed the _ValidateNotNullOrEmpty_ check from the NugetApiKey parameter and implemented `if` statements checking for the parameter in Publish-Module and Publish-PSArtifactUtility when constructing parameters, in the same way credentials are added conditionally.

I ran tests on my machine after making the changes and got the following results:
```powershell
Tests Passed: 348, Failed: 10, Skipped: 53
```
Most of the failures were related to the dotnet cli tool, I have version 3.0.0 preview installed and imagine that might be conflicting.

With this change the cmdlet fails if no apikey is passed, in the same way as if an invalid apikey is passed, with a 403 status code from the API.


src/PowerShellGet/private/functions/Publish-PSArtifactUtility.ps1
```diff
@@ -531,16 +530,21 @@ $CsprojContent = @"
...
         elseif($script:NuGetExePath) {
             $StartProcess_params['FilePath'] = $script:NuGetExePath
 
             $ArgumentList = @('push')
             $ArgumentList += "`"$NupkgPath`""
             $ArgumentList += @('-source', "`"$($Destination.TrimEnd('\'))`"")
-            $ArgumentList += @('-apikey', "`"$NugetApiKey`"")
             $ArgumentList += '-NonInteractive'
+            if ($PSBoundParameters.Containskey('NugetApiKey')) {
+                $ArgumentList += @('--api-key', "`"$NugetApiKey`"")
```
edyoung, line 546: Nuget.exe doesn't have an argument --api-key, stick with -apikey

edyoung (changes requested):
@heidarj thanks! I think this is a fine approach. Please see my comment, and check that the error message you get when you try to publish to psgallery without an apikey is understandable.

edyoung:
@heidarj I'm pretty sure the argument to nuget.exe I call out above is wrong, but otherwise this should be fine. Let us know if you have time to update?

heidarj:
@edyoung I implemented the localized _NuGetApiKeyIsRequiredForNuGetBasedGalleryService_ error message for when an API key is not specified so that it now throws the same error as before, if the repo you are pushing to requires an API key. Please let me know if any further changes are needed, otherwise I am good with this as is.

edyoung:
Thanks! I think it's still a problem to pass --api-key when calling nuget, because it doesn't support the -- based parameter form. Also looks like the branch needs to be rebased or merged to match what's currently in development. If it's possible to fix that will happily merge.

PR #469, opened 2019-04-30
update tests 

edyoung:
refactoring broke prerelease tests, since 1.0.0-alpha1 is not a valid [version]


alerickson (approved):

edyoung:
AcceptLicense tests were failing due to #472. I updated the module in poshtestgallery and expect it to pass now.

PR #471, opened 2019-05-01
Added scope to Update-Module. Added localized error message

lwajswaj:
This PR adds support for -Scope parameter for Update-Module; allowing to overwrite built-in logic to support updatings Modules in system-path.

Fixed issues #468


edyoung:
looks good to me. thanks!

PR #476, opened 2019-05-13
Fix for Get-Package bug in OSX

alerickson:


edyoung:
so the claim is that the change in Get-ChildItem behavior is deliberate? If so, ok.  BTW the fix in https://github.com/PowerShell/PowerShellGet/pull/395 was the exact opposite - remove -Force so that hidden files were not published

alerickson:
Resolves https://github.com/OneGet/oneget/issues/450 -- issue due to a behavior change in .NET Core 3.0

Issue #340, opened 2018-10-03
Proposal: Allow multiple repositories to contain the same package


edyoung:
If you have 2 repositories registered, and both of them contain a package Foo, you get an error when you run Install-Module Foo, saying that you have to pick a repository.

It would be better IMHO if this was permitted, and PowerShellGet automatically picked a repository when this happened. The main question is: when 2 repositories contain the requested module, how are ties broken - ie, which one do we pick? 

## Suggestion:

When Install-Module is run, we find all candidate repositories which contain a matching module, as at present. Then we order them using the method below to pick the preferred repo.

First, trusted vs untrusted is used. All trusted repositories are automatically higher ranked than all untrusted repos. 

Second, introduce an explicit ordering for repositories, and a way for users to reorder them. The PSRepositoryInfo flag would have a Priority property added, which would be an int value, with 0 as lowest priority (last to be picked). This can be null, if not explicitly configured, which is treated as 0. Priorities need not be unique or consecutive.

Users can set the Priority property at Register-PSRepository time, and with Set-PSRepository -Priority. If not specified they get 0 priority.

Third, if 2 repositories have the same trust status and the same priority (including null) we pick them in alphabetical order.

Get-PSRepository would display the repos in priority order, NOT the order of the ordered table they are stored in (which is sorted by key). Additionally, it would show the Priority field by default.

This should not be a breaking change, since it only updates behavior for situations which would previously have caused an error.

## Questions:
What if a repo with lower priority contains a module with higher version? Should that be taken into account? I think no.

Is it confusing to display an untrusted repo with Priority 999 and a trusted one with priority 1, and then choose the trusted one first?

Should the behavior of any other commands change?

What about package management, any changes needed there?

alerickson:
I think this would be a nice addition and certainly make for a smoother experience.
It makes more sense to me to sort by priority, then trust, then alphabetically.  If someone values priorities enough to explicitly set them for their repositories it seems like that should take precedence over whether the repo is trusted or not.

I don't think any other PSGet commands would change, but I'd have to look into package management.

SydneyhSmith:
I agree that this is a good improvement to make. Something interesting when looking at other commands that could be effective is that Update-Module and Uninstall-Module work even with a module in multiple repositories i.e. they already use a hierarchy (based on version I think for Uninstall) while this wouldn't break Uninstall, it might make sense to extend the notion of priority of registers for consistency across commands

edyoung:
[...]

PR #481, opened 2019-05-21
Issue 340 - Prefer Trusted Repositories

awickham10:
This addresses a portion of issue #340.

* Get-PSRepository now returns trusted repos, then untrusted repos. They are also sorted alphabetically. 
* Install-Module now does a Find-Module based on the parameters and if there are multiple repositories found, the first result is picked and set as the Source parameter for Install-Package. 

Tests were added for both changes. I'm concerned that Find-Module will have a performance impact, but I'm not sure how else to do that check unless it's done at the PackageManagement\Install-Package function instead.


alerickson:
Thanks for putting in this PR!  I’ll take a look at it within the next couple days.  Just wanted to give you an update

awickham10:
> Thanks for putting in this PR! I’ll take a look at it within the next couple days. Just wanted to give you an update

No rush, I'm obviously still working through some kinks 😅

PR #487, opened 2019-05-30
Add PassThru parameter to Install/Update functions

### LINKED ISSUE #14: Install-Module should have a -Passthru parameter
This is useful in atleast following 2 scenario
- Allows to use the installed module in the pipeline (Think Install-Module | Get-DscResource as one example)
- Allows the scripters to notify the end user that module got installed and its location

Workaround is an extra call to Get-Module -ListAvailable

ThomasNieto:
This PR closes #14. The -PassThru parameter outputs a `Microsoft.PowerShell.Commands.PSRepositoryItemInfo` object.


ThomasNieto:
@alerickson Is there a way to trigger a rebuild of just the job that failed? Also, I'm not sure why it failed in the first place since it passed on all the other platforms and the last run. The test that failed is not part of the changes I made.

alerickson:
@tnieto88 everything's solid!  The only test that failed is a notoriously flakey.  I'm rerunning the failed test image but the PR is fine

ThomasNieto:
@alerickson thanks! It looks like that test passed now.

PR #496, opened 2019-06-12
Publish error

alerickson:


tools/build.psm1
```diff
@@ -120,7 +124,7 @@ function Install-PackageManagement {
         $null = Microsoft.PowerShell.Management\New-Item -Path $TempModulePath -Force -ItemType Directory
         $OneGetModuleName = 'PackageManagement'
         try {
-            & $NugetExeFilePath install $OneGetModuleName -source https://www.powershellgallery.com/api/v2 -outputDirectory $TempModulePath -verbosity detailed
+            & $NugetExeFilePath install $OneGetModuleName -source https: / / www.powershellgallery.com / api / v2 -outputDirectory $TempModulePath -verbosity detailed
```
edyoung: looks like there are spaces around the / - I don't think we want that. If you quote the URLs the autoformatter won't do that.

src/PowerShellGet/private/functions/Publish-NugetPackage.ps1
```diff
@@ -26,7 +26,6 @@ function Publish-NugetPackage {
         $ArgumentList += "`"$NupkgPath`""
         $ArgumentList += @('-source', "`"$Destination`"")
         $ArgumentList += @('-apikey', "`"$NugetApiKey`"")
-        $ArgumentList += '-NonInteractive'
```
edyoung, line 29: So, we do need nuget.exe to be interactive? That seems weird and surprising, do you know why?

alerickson: I'm actually going to investigate another fix.  I'm not sure why this works, but I'd rather leave this as non-interactive.

PR #503, opened 2019-06-17
Condense release notes due to 10k char limit

alerickson:


src/PowerShellGet/PowerShellGet.psd1
```diff
@@ -237,45 +237,21 @@ New features
...
 
 * Initial release from GitHub.
 * PowerShellCore support.
-* Security enhancements including the enforcement of catalog-signed modules during installation.
-* Authenticated Repository support.
-* Proxy Authentication support.
-* Responses to a number of user requests and issues.
+
+## For full history of release notes see changelog
 '@
```
SydneyhSmith, line 256: Should we also include where to view the changelog or a link to it?

SydneyhSmith (approved):

src/PowerShellGet/PowerShellGet.psd1
alerickson: Sounds good, just added that

PR #521, opened 2019-08-08
Fix DscResources foldername

TravisEz13:
DscResources foldername is cases sensitive on platforms that have case sensitive file systems


johlju:
@TravisEz13 Maybe we should update the folder structure https://docs.microsoft.com/en-us/powershell/dsc/resources/authoringresourcemof#folder-structure-for-a-mof-resource with a note that is is case sensitive on on platforms that have case sensitive file system? What looks for `DscResource` and not `DSCResource`, maybe that can be added as well?

johlju (approved):
LGTM

TravisEz13:
@johlju  I'm not sure a final decision has been made if this is a bug or by design.  Please work with the Nitin and @amitsara  and they agreed they would own everything beyond this change wrt to this case sensitivity.

PR #558, opened 2019-12-11
Improves discovery of usable nuget.exe binary

### LINKED ISSUE #556: Issues when older nuget.exe installed in Program Files
## Issue Description:
Publish-Module favors any `nuget.exe` file found in Program Files, even if it doesn't meet the minimum version requirement.

I am experiencing this, as a non-admin user on my desktop, an older version of `nuget.exe` is installed in Program Files. But if I download a newer version from Microsoft and place it in my `AppData Local` path, the logic does not allow it to be discovered, but instead prompts me to install `nuget.exe` in one of the following locations:

1. Program Files
2. User's App Data local path
3. A location available from $env:PATH

Here is the location that this is happening:
https://github.com/PowerShell/PowerShellGet/blob/712d6d3b64980bba1a46f20e5e843aef8319de53/src/PowerShellGet/private/functions/Install-NuGetClientBinaries.ps1#L96

### Recommended fix:
``` Powershell
if( (Microsoft.PowerShell.Management\Test-Path -Path $programDataExePath) -and ($programDataExePath | Microsoft.PowerShell.Core\Get-Command).Version -ge $script:NuGetExeMinRequiredVersion )
```

https://github.com/PowerShell/PowerShellGet/blob/712d6d3b64980bba1a46f20e5e843aef8319de53/src/PowerShellGet/private/functions/Install-NuGetClientBinaries.ps1#L100

### Recommended fix:
``` powershell
elseif( (Microsoft.PowerShell.Management\Test-Path -Path $applocalDataExePath) -and ($applocalDataExePath | Microsoft.PowerShell.Core\Get-Command).Version -ge $script:NuGetExeMinRequiredVersion )
```

bwright86:
Includes minimum required version to the checks in Program Files and User AppData Local folders. This prevents issues with outdated nuget.exe binaries in Program Files occluding newer versions in User's AppData folders or locations on the PATH.

Would resolve #556

### DIFF
--- a/src/PowerShellGet/private/functions/Install-NuGetClientBinaries.ps1
+++ b/src/PowerShellGet/private/functions/Install-NuGetClientBinaries.ps1
@@ -93,11 +93,13 @@ function Install-NuGetClientBinaries
             $applocalDataExePath = Microsoft.PowerShell.Management\Join-Path -Path $script:PSGetAppLocalPath -ChildPath $script:NuGetExeName
 
             # Check if NuGet.exe is available under one of the predefined PowerShellGet locations under ProgramData or LocalAppData
-            if(Microsoft.PowerShell.Management\Test-Path -Path $programDataExePath)
+            if( (Microsoft.PowerShell.Management\Test-Path -Path $programDataExePath) -and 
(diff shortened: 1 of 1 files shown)


alerickson:
Thanks for the reporting the issue and getting a fix in so quickly!

Issue #510, opened 2019-07-05
New-NugetPackage fails on german dotnet core (win)


jmarkert:
Steps to reproduce
------------------
Publishing a module to my private PSRepositoy (Nexus3) fails. I'm not sure, but I guess it is broken since I have upgraded PScore from 6.1 to 6.2

```powershell
Publish-Module -Path ".\dest\$ModuleName\$version\" -Repository $MyPrivateRepo -NuGetApiKey $NuGetApiKey -Credential $Credential
```

Actual behavior
---------------

```none
Publish-PSArtifactUtility : Failed to generate the compressed file for module 'Cannot index into a null array.'.
At C:\program files\powershell\6\Modules\PowerShellGet\PSModule.psm1:10785 char:17
+ ...             Publish-PSArtifactUtility @PublishPSArtifactUtility_Param ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidOperation: (:) [Write-Error], WriteErrorException
+ FullyQualifiedErrorId : FailedToCreateCompressedModule,Publish-PSArtifactUtility
```
I debugged the process and found the problem in the New-NugetPackage function
https://github.com/PowerShell/PowerShellGet/blob/7de99ee0c38611556e5c583ffaca98bb1922a0d4/src/PowerShellGet/private/functions/New-NugetPackage.ps1#L119

dotnet.exe returns and localized string, so the -match does not work. For the positive match in german the expression should be something like:
```none
'Das Paket "(.*.nupkg)" wurde erfolgreich erstellt.' 
```
For testing I have changed the expression and the package creation gets recognized correctly. 

Environment data
----------------
I'am on german Windows 10 1903

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.2.1
PSEdition                      Core
GitCommitId                    6.2.1
OS                             Microsoft Windows 10.0.18362
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

```powershell
> Get-Module -ListAvailable PowerShellGet,PackageManagement
    Directory: C:\program files\powershell\6\Modules

ModuleType Version    Name                                PSEdition ExportedCommands
[...]

PR #567, opened 2019-12-25
Fixing Publish Failure (#510) 

KUTlime:
Publish fails on poor searching of a nupkg file.
The issue is reported [here](https://github.com/PowerShell/PowerShellGet/issues/510).
On non en-US environments, MSBuild produces culture variant message. `New-NugetPackage` relies on the English string.

obrazek

The main problem here is a wrong perspective. 
We are dealing with a file path as a `string` not as a `FileIO` instance.


anamnavi:
Closing in favor of similar changes merged in via PR #613

KUTlime:
@anamnavi Nice, thank you for information. It took only 3 years. 🤣

PR #598, opened 2020-04-15
Add enforcement of TLS1.2

alerickson:
Due to the PowerShell Gallery needing to enforce TLS 1.2, we've had a number of complaints from those still using lower versions.  Additions in this PR will save the user's current security protocol settings, then change the security protocol to TLS 1.2 in order to send requests via the updated version, then revert the security protocol back to the user's setting before the request was sent.

### DIFF
--- a/src/PowerShellGet/public/psgetfunctions/Find-Module.ps1
+++ b/src/PowerShellGet/public/psgetfunctions/Find-Module.ps1
@@ -90,6 +90,10 @@ function Find-Module {
     )
 
     Begin {
+        # Change security protocol to TLS 1.2
+        $script:securityProtocol = [Net.ServicePointManager]::SecurityProtocol
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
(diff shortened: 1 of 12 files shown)


src/PowerShellGet/public/psgetfunctions/Find-Module.ps1
```diff
@@ -90,6 +90,10 @@ function Find-Module {
     )
 
     Begin {
+        # Change security protocol to TLS 1.2
+        $script:securityProtocol = [Net.ServicePointManager]::SecurityProtocol
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```
bergmeister, line 95: I don't think it is a good idea @alerickson to hard-code the usage of just one protocol. If the Gallery wants to enforce `Tls13` in the future, then that change would brick all versions of PowerShellGet where the protocol is hard-coded like that and one would not even be able to change it on a higher level. cc @SteveL-MSFT 
Rather, I suggest to _add_ the `Tls12` security protocol as follows, which is the standard, recommended practice:
```pwsh
 [Net.ServicePointManager]::SecurityProtocol =  [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
```

PR #601, opened 2020-04-22
Remove catalog file and up version number

alerickson:
Catalog file is a relic from older days when it was used a fix for a Windows component.  It's no longer adding any value and can causes problems with installations.


bergmeister:
@alerickson To me, it looks like, `PowerShellGet` rather needs to correctly publish the catalog file so that things like that do not happen:
https://twitter.com/Bu5yGiraffe/status/1250956672127963139/photo/1

PR #683, opened 2022-10-17
Revert "Change all instances of 'whitelist' to 'allowlist'"

xtqqczze:
This reverts commit cd09bb671aa892e13ca5ad71dfb7c4d49ab38c80.

cc: @alerickson, @TravisEz13, @adityapatwardhan

It is confusing to change instances of whitelist in comments and changelog, given that the term is referring to the following PowerShell variable:

https://github.com/PowerShell/PowerShellGetv2/blob/7ef9df4738fbe2eefb144ff94b8941da0ccdec7f/src/PowerShellGet/private/modulefile/PartOne.ps1#L320-L323

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -157,7 +157,7 @@ Bug fixes
 New features
 
 - Allow Pester/PSReadline installation when signed by non-Microsoft certificate (#258)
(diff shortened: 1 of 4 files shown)


xtqqczze:
Google recommends against using the terms: allowlist (verb), allowlisted, allowlisting, see:

https://developers.google.com/style/word-list#allowlist

